PR #45, opened 2026-03-13
[#2] Supabase schema migration and client helper

### LINKED ISSUE #2: [P0-2] Supabase Schema
## Context
Set up the database schema from the proposal. Three tables: `storylines`, `plots`, `donations`. The schema includes deduplication constraints (`UNIQUE(tx_hash, log_index)`), foreign keys, and an indexer-set `writer_type` column. Also add a `hidden` boolean column (default false) to `storylines` and `plots` for MVP content moderation (§8 Risks table).

## Sub-tickets
- [ ] **P0-2a**: Create SQL migration file(s) with all three tables exactly as specified in §4.1, plus the `hidden` column on `storylines` and `plots`.
- [ ] **P0-2b**: Create Supabase client helper (`lib/supabase.ts`) — server-side client using service role key, and a browser-side client using anon key. Follow the pattern used in other project repos (dropcast, reviewme-fun).

## Proposal Reference
§4.1 (Supabase Schema)

realproject7:
## Summary
- **P0-2a**: SQL migration (`supabase/migrations/00001_schema.sql`) with `storylines`, `plots`, and `donations` tables per §4.1
  - `UNIQUE(tx_hash, log_index)` deduplication on all three tables
  - Foreign keys from `plots` and `donations` to `storylines`
  - `writer_type` (smallint, default 0) for indexer-set agent detection
  - `hidden` (boolean, default false) on `storylines` and `plots` for MVP moderation (§8)
- **P0-2b**: `lib/supabase.ts` with browser-side (anon key) and server-side (service role) clients, following the dropcast pattern
  - Full TypeScript `Database` interface with Row/Insert/Update types for all three tables
  - Convenience type aliases: `Storyline`, `Plot`, `Donation`

Fixes #2

## Test plan
- [ ] Verify SQL migration runs cleanly against a fresh Supabase project
- [ ] Verify `lib/supabase.ts` type-checks (CI confirms this)
- [ ] Confirm schema matches §4.1 requirements: deduplication, FKs, writer_type, hidden columns

### DIFF
--- a/lib/supabase.ts
+++ b/lib/supabase.ts
@@ -1,0 +1,173 @@
+import { createClient, SupabaseClient } from "@supabase/supabase-js";
+
+const supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL || "";
+const supabaseAnonKey = process.env.NEXT_PUBLIC_SUPABASE_ANON_KEY || "";
+
+// Browser-side client (anon key, respects RLS)
+export const supabase: SupabaseClient<Database> | null =
+  supabaseUrl && supabaseAnonKey
+    ? createClient<Database>(supabaseUrl, supabaseAnonKey)
+    : null;
+
+// Server-side client (service role, bypasses RLS)
+export function createServerClient(): SupabaseClient<Database> | null {
+  const serviceRoleKey = process.env.SUPABASE_SERVICE_ROLE_KEY || "";
+  if (!supabaseUrl || !serviceRoleKey) return null;
+
+  return createClient<Database>(supabaseUrl, serviceRoleKey, {
+    auth: {
+      autoRefreshToken: false,
+      persistSession: false,
+    },
+  });
+}
+
+// ---------------------------------------------------------------------------
+// Database types
+// ---------------------------------------------------------------------------
+
+export type Json =
+  | string
+  | number
+  | boolean
+  | null
+  | { [key: string]: Json | undefined }
+  | Json[];
+
+export interface Database {
+  public: {
+    Tables: {
+      storylines: {
+        Row: {
+          id: number;
+          storyline_id: number;
+          writer_address: string;
+          token_address: string;
+          title: string;
+          plot_count: number;
+          last_plot_time: string | null;
+          has_deadline: boolean;
+          sunset: boolean;
+          writer_type: number;
+          hidden: boolean;
+          tx_hash: string;
+          log_index: number;
+          block_timestamp: string | null;
+          indexed_at: string;
+        };
+        Insert: {
+          id?: never;
+          storyline_id: number;
+          writer_address: string;
+          token_address: string;
+          title: string;
+          plot_count?: number;
+          last_plot_time?: string | null;
+          has_deadline?: boolean;
+          sunset?: boolean;
+          writer_type?: number;
+          hidden?: boolean;
+          tx_hash: string;
+          log_index: number;
+          block_timestamp?: string | null;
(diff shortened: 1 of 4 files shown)


project7-interns (approved):
**T2b review: APPROVE**

Schema review against issue #2 requirements:

✅ Three tables (`storylines`, `plots`, `donations`) — all present
✅ `UNIQUE(tx_hash, log_index)` deduplication constraints on all three tables
✅ Foreign keys: `plots.storyline_id → storylines(id)`, `donations.storyline_id → storylines(id)`
✅ `writer_type smallint not null default 0` on `storylines` — indexer-set, well-commented
✅ `hidden boolean not null default false` on `storylines` and `plots` (not on `donations`) — matches §8
✅ Useful indexes on `writer_address` and `storyline_id` FKs
✅ `amount text` for wei — avoids precision loss, correct choice

**lib/supabase.ts:**
✅ Browser client (anon key, respects RLS) with null fallback when env vars missing
✅ Server client (service role, bypasses RLS) with auth disabled — correct pattern
✅ Full `Database` interface with Row/Insert/Update types matching SQL exactly
✅ Convenience type aliases exported

**Minor note (non-blocking):** `plots.storyline_id` references the surrogate PK `storylines.id`, not the on-chain `storylines.storyline_id`. The naming overlap could confuse future contributors — consider renaming the FK column to `storyline_fk` or adding a code comment in a future PR. Not blocking since the relational logic is correct.

Clean, spec-compliant implementation.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR is close, but the schema currently bakes in one silent data-classification bug and one relationship mismatch that will make the indexer harder to implement correctly.

### Findings
- [high] `writer_type` is supposed to be indexer-set, but the migration gives it `default 0`
  - File: `supabase/migrations/00001_schema.sql:16`
  - Suggestion: remove the default and require inserts to provide `writer_type` explicitly, so rows cannot silently land as `human` when the indexer forgets or fails to set it.

- [high] Child `storyline_id` foreign keys point at the surrogate row id instead of the storyline identifier
  - File: `supabase/migrations/00001_schema.sql:31`
  - Suggestion: align the relationship so `plots.storyline_id` and `donations.storyline_id` reference the canonical storyline identifier defined in the schema/proposal, rather than `storylines.id`. As written, a chain event carrying `storyline_id` cannot be inserted directly without an extra lookup/translation step, and the column name becomes misleading.

### Decision
Request changes. The current migration risks silently misclassifying writer type and creates a storyline relationship that does not match the event-facing identifier implied by the schema.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit addresses the two schema correctness issues from the prior review. `writer_type` no longer silently defaults to human, and the child tables now reference the canonical on-chain `storyline_id` directly.

### Findings
- None.

### Decision
Approve. The migration and Supabase helper now match the issue scope, and the PR checks are green.

project7-interns (approved):
**T2b re-review: APPROVE**

Both fixes verified:

1. ✅ `writer_type` → nullable `smallint` (no default). NULL = unclassified, 0 = human, 1 = agent. No more silent classification. TS types updated to `number | null`.
2. ✅ FKs on `plots`/`donations` now reference `storylines(storyline_id)` directly, backed by new `storylines_onchain_unique` constraint. Indexer can insert with on-chain ID — no surrogate lookup needed. Resolves my earlier non-blocking note as well.

Clean fix, no regressions.

Issue #8, opened 2026-03-13
[P1-3] Contract ABI & Constants


realproject7:
## Context
Define the event ABIs and contract addresses so the indexer can parse onchain events. Contracts aren't deployed yet, but the event signatures are finalized in §4.1 and §4.3.

## Sub-tickets
- [ ] **P1-3a**: Create `lib/contracts/abi.ts` — define ABIs for `PlotChained`, `StorylineCreated`, and `Donation` events. Include the `chainPlot` and `createStoryline` function ABIs for frontend contract calls.
- [ ] **P1-3b**: Create `lib/contracts/constants.ts` — placeholder addresses for StoryFactory, $PLOT token. Include the known addresses from §12 (MCV2_Bond: `0xc5a076cad94176c2996B32d8466Be1cE757FAa27`, MCV2_BondPeriphery: `0x492C412369Db76C9cdD9939e6C521579301473a3`, Uniswap V4 Router, Permit2, ERC-8004 registry: `0x8004A169FB4a3325136EB29fA0ceB6D2e539a432`).

## Proposal Reference
§4.1 (Contract events), §4.3 (StoryFactory), §12 (External Dependencies)

PR #49, opened 2026-03-13
[#9] Inline Indexer — Plots

### LINKED ISSUE #9: [P1-4] Inline Indexer — Plots
## Context
The primary indexing path: frontend calls this API after a `chainPlot()` transaction confirms. Inline indexer pattern (§4.1) — no cron, no event listener. Frontend triggers indexing by POSTing the txHash.

The indexer fetches the receipt, parses the PlotChained event, fetches content from IPFS by CID, verifies the content hash matches the onchain contentHash, and upserts to Supabase. If IPFS fetch times out (10s), it falls back to content provided in the request body.

**Important**: Use `publicClient.getBlock()` to get the block timestamp — it is NOT available on the transaction receipt in viem.

## Sub-tickets
- [ ] **P1-4a**: Implement `POST /api/index/plot` route with the full flow: fetch receipt → parse PlotChained event → fetch content from IPFS (10s timeout via `AbortSignal.timeout`) → fallback to request body content if IPFS fails → verify `keccak256(content) == onchain contentHash` → get block timestamp via `getBlock()` → upsert to Supabase. Deduplicate on `(tx_hash, log_index)`.

## Proposal Reference
§4.1 (Indexing Layer: Supabase Inline Indexer)

## Dependencies
P1-1 (Filebase helper), P1-2 (content utils), P1-3 (ABI definitions), P0-2 (Supabase schema)

realproject7:
## Summary
- **P1-4a**: `src/app/api/index/plot/route.ts` — POST endpoint implementing the full inline indexer flow:
  1. Fetch transaction receipt
  2. Decode `PlotChained` event from StoryFactory logs
  3. Fetch content from IPFS gateway (10s timeout via `AbortSignal.timeout`)
  4. Fallback to request body content if IPFS fetch fails
  5. Verify `keccak256(content) == onchain contentHash`
  6. Get block timestamp via `getBlock()` (not available on receipt)
  7. Upsert to Supabase with `(tx_hash, log_index)` deduplication
- `supabase/migrations/00002_plots_content_column.sql` — adds `content TEXT` column to plots table (proposal §4.1 primary read path)
- `lib/supabase.ts` — updated Database types to include `content` field on plots
- `lib/viem.ts` — Base public client helper (configurable RPC URL)

Fixes #9

## Test plan
- [x] `tsc --noEmit` passes
- [x] `vitest run` — 22/22 passing (existing tests unaffected)
- [ ] Integration test on Base Sepolia with a real `chainPlot` transaction
- [ ] Verify migration applies cleanly on Supabase

### DIFF
--- a/lib/supabase.ts
+++ b/lib/supabase.ts
@@ -96,6 +96,7 @@ export interface Database {
           storyline_id: number;
           plot_index: number;
           writer_address: string;
+          content: string | null;
           content_cid: string;
           content_hash: string;
           hidden: boolean;
@@ -109,6 +110,7 @@ export interface Database {
           storyline_id: number;
           plot_index: number;
           writer_address: string;
+          content?: string | null;
           content_cid: string;
           content_hash: string;
           hidden?: boolean;
@@ -122,6 +124,7 @@ export interface Database {
           storyline_id?: number;
           plot_index?: number;
           writer_address?: string;
+          content?: string | null;
           content_cid?: string;
           content_hash?: string;
           hidden?: boolean;
--- a/lib/viem.ts
+++ b/lib/viem.ts
@@ -8,5 +8,5 @@ import { base } from "viem/chains";
  */
 export const publicClient = createPublicClient({
   chain: base,
-  transport: http(process.env.NEXT_PUBLIC_BASE_RPC_URL || undefined),
+  transport: http(process.env.BASE_RPC_URL || undefined),
 });
--- a/lib/viem.ts
+++ b/lib/viem.ts
@@ -4,7 +4,7 @@ import { base } from "viem/chains";
 /**
  * Public client for reading from Base.
  *
- * Uses the default Base RPC. Override via NEXT_PUBLIC_BASE_RPC_URL env var.
+ * Uses the default Base RPC. Override via BASE_RPC_URL env var.
  */
 export const publicClient = createPublicClient({
   chain: base,
--- a/lib/viem.ts
+++ b/lib/viem.ts
@@ -1,0 +1,12 @@
+import { createPublicClient, http } from "viem";
+import { base } from "viem/chains";
+
+/**
+ * Public client for reading from Base.
+ *
+ * Uses the default Base RPC. Override via NEXT_PUBLIC_BASE_RPC_URL env var.
+ */
+export const publicClient = createPublicClient({
+  chain: base,
+  transport: http(process.env.NEXT_PUBLIC_BASE_RPC_URL || undefined),
+});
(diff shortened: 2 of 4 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The route structure matches issue #9, but the current event-selection logic makes the endpoint fail until a real StoryFactory address is manually filled in. That is a functional blocker for the inline indexer path this ticket is supposed to deliver.

### Findings
- [high] The route filters receipt logs by `STORY_FACTORY`, but that constant is still the zero-address placeholder from issue #8, so `/api/index/plot` cannot find a real `PlotChained` log and will return `"PlotChained event not found in receipt"` for valid transactions.
  - File: `src/app/api/index/plot/route.ts:36`
  - Suggestion: select the log by decoding receipt logs against the ABI and picking `eventName === "PlotChained"`, or otherwise make the contract address injectable instead of hard-wiring the undeployed placeholder constant here.

### Decision
Request changes because the current implementation does not provide a usable inline indexer flow in environments where PlotLink contracts are not yet deployed, even though the rest of the route is wired correctly.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The blocking issue from the previous review is fixed. The route now identifies `PlotChained` logs by event signature rather than the undeployed placeholder contract address, which makes the inline indexer path usable for valid receipts.

### Findings
- None.

### Decision
Approve because the prior functional blocker is resolved, the route still matches issue #9 scope, and the updated CI check passed.

project7-interns (approved):
## T2b Review: APPROVE

All three requested changes confirmed:
1. **txHash validation** — regex `/^0x[0-9a-fA-F]{64}$/` applied before RPC call
2. **Migration idempotency** — `IF NOT EXISTS` on ADD COLUMN
3. **Env var rename** — code correctly uses `BASE_RPC_URL`

Nit: JSDoc comment in `lib/viem.ts` still references old `NEXT_PUBLIC_BASE_RPC_URL` name — cosmetic only, non-blocking.

Event signature matching via `topics[0]` is correct. Hash verification, IPFS fallback, and Supabase upsert all look good. Solid first indexer endpoint.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up fixes are correct: tx hash validation is in place, the migration is idempotent, and the RPC env var is server-only now. The updated head commit remains within issue #9 scope and CI passed.

### Findings
- None.

### Decision
Approve because the additional changes address the remaining review concerns without introducing a new correctness issue.

PR #53, opened 2026-03-13
[#52] Fix StorylineCreated ABI + genesis plot indexing

### LINKED ISSUE #52: [BUG] Storyline indexer missing genesis plot + StorylineCreated ABI incomplete
## Bug Description

Two related issues found during code review of PRs #45-#50:

### 1. StorylineCreated event ABI missing fields (`lib/contracts/abi.ts`)

The `StorylineCreated` event is missing `openingCID` and `openingHash` fields. Per proposal §4.1, `createStoryline()` emits these so the indexer can fetch and verify the genesis plot content.

**Current:**
```typescript
export const storylineCreatedEvent = {
  inputs: [
    { name: "storylineId", type: "uint256", indexed: true },
    { name: "writer", type: "address", indexed: true },
    { name: "tokenAddress", type: "address", indexed: false },
    { name: "title", type: "string", indexed: false },
    { name: "hasDeadline", type: "bool", indexed: false },
  ],
} as const;
```

**Expected — add these two fields:**
```typescript
{ name: "openingCID", type: "string", indexed: false },
{ name: "openingHash", type: "bytes32", indexed: false },
```

### 2. Storyline indexer does not store genesis plot (`src/app/api/index/storyline/route.ts`)

The storyline indexer creates a `storylines` row with `plot_count: 1` but **never inserts a row into the `plots` table** for the opening chapter (plot_index = 0). This means readers cannot see the genesis plot.

**Fix:** After creating the storyline record, the indexer must:
1. Read `openingCID` and `openingHash` from the decoded event (requires ABI fix above)
2. Fetch content from IPFS via `https://<bucket>.s3.filebase.com/plotlink/<openingCID>`
3. Fall back to request body `content` field if IPFS fetch fails
4. Verify `keccak256(toHex(content)) === openingHash`
5. Insert into `plots` table with `plot_index: 0`, `storyline_id`, `writer_address`, `content`, `content_cid: openingCID`, `content_hash: openingHash`

Reference the existing plot indexer (`src/app/api/index/plot/route.ts`) for the IPFS fetch + hash verify + Supabase upsert pattern.

### 3. `createStoryline` function ABI may also need `openingCID` and `openingHash` params

Check `lib/contracts/abi.ts` — if the `createStoryline` function ABI is also missing these input params, add them to match proposal §4.1.

## Acceptance Criteria

- [ ] `storylineCreatedEvent` in `lib/contracts/abi.ts` includes `openingCID` (string) and `openingHash` (bytes32)
- [ ] `createStoryline` function ABI in `lib/contracts/abi.ts` includes `openingCID` and `openingHash` params if missing
- [ ] Storyline indexer inserts a `plots` row for the genesis plot (plot_index = 0)
- [ ] Genesis plot content is fetched from IPFS, verified against hash, stored in `content` column
- [ ] Falls back to request body if IPFS fetch fails (same pattern as plot indexer)

realproject7:
## Summary
- **ABI fix**: Added missing `openingCID` (string) and `openingHash` (bytes32) fields to `storylineCreatedEvent` in `lib/contracts/abi.ts`
- **Genesis plot indexing**: Storyline indexer now inserts a `plots` row for plot_index=0 after creating the storyline record:
  - Fetches content from IPFS via openingCID (10s timeout)
  - Falls back to request body `content` field
  - Verifies keccak256 hash match against openingHash
  - Upserts to plots table with deduplication
- `createStorylineFunction` ABI already had the correct params (no change needed)

Fixes #52

## Test plan
- [x] `tsc --noEmit` passes
- [x] `vitest run` — 22/22 passing
- [ ] Verify StorylineCreated topic0 changed (ABI fields affect signature)
- [ ] Integration test: createStoryline tx → verify both storylines and plots rows created

### DIFF
--- a/lib/contracts/abi.ts
+++ b/lib/contracts/abi.ts
@@ -30,6 +30,8 @@ export const storylineCreatedEvent = {
     { name: "tokenAddress", type: "address", indexed: false },
     { name: "title", type: "string", indexed: false },
     { name: "hasDeadline", type: "bool", indexed: false },
+    { name: "openingCID", type: "string", indexed: false },
+    { name: "openingHash", type: "bytes32", indexed: false },
   ],
 } as const;
 
--- a/src/app/api/index/storyline/route.ts
+++ b/src/app/api/index/storyline/route.ts
@@ -7,7 +7,11 @@ import {
   storylineCreatedEvent,
 } from "../../../../../lib/contracts/abi";
 import { detectWriterType } from "../../../../../lib/contracts/erc8004";
+import { hashContent } from "../../../../../lib/content";
 import type { Database } from "../../../../../lib/supabase";
+
+const IPFS_GATEWAY = "https://ipfs.filebase.io/ipfs/";
+const IPFS_TIMEOUT_MS = 10_000;
 
 /** StorylineCreated event topic0 */
 const STORYLINE_CREATED_TOPIC = encodeEventTopics({
@@ -22,6 +26,7 @@ function error(message: string, status = 400) {
 export async function POST(req: Request) {
   const body = await req.json();
   const txHash = body.txHash as Hex | undefined;
+  const fallbackContent = body.content as string | undefined;
 
   if (!txHash || !/^0x[0-9a-fA-F]{64}$/.test(txHash)) {
     return error("Missing or invalid txHash");
@@ -64,8 +69,15 @@ export async function POST(req: Request) {
     return error("Unexpected event type");
   }
 
-  const { storylineId, writer, tokenAddress, title, hasDeadline } =
-    decoded.args;
+  const {
+    storylineId,
+    writer,
+    tokenAddress,
+    title,
+    hasDeadline,
+    openingCID,
+    openingHash,
+  } = decoded.args;
 
   // 4. Get block timestamp
   let blockTimestamp: bigint;
@@ -81,13 +93,37 @@ export async function POST(req: Request) {
   // 5. Detect writer type via ERC-8004 (best-effort, defaults to human)
   const writerType = await detectWriterType(writer);
 
-  // 6. Upsert to Supabase
+  // 6. Fetch genesis plot content from IPFS (with fallback)
+  let genesisContent: string | null = null;
+  try {
+    const ipfsRes = await fetch(`${IPFS_GATEWAY}${openingCID}`, {
+      signal: AbortSignal.timeout(IPFS_TIMEOUT_MS),
+    });
+    if (!ipfsRes.ok) throw new Error(`IPFS status ${ipfsRes.status}`);
+    genesisContent = await ipfsRes.text();
+  } catch {
+    if (fallbackContent) {
+      genesisContent = fallbackContent;
+    }
+  }
+
+  // 7. Verify genesis content hash (if content was fetched)
+  if (genesisContent !== null) {
+    const computedHash = hashContent(genesisContent);
+    if (computedHash !== openingHash) {
+      genesisContent = null; // reject tampered content
+    }
+  }
+
+  // 8. Upsert storyline to Supabase
   const supabase = createServerClient();
   if (!supabase) {
     return error("Supabase not configured", 500);
   }
 
-  const row: Database["public"]["Tables"]["storylines"]["Insert"] = {
+  const timestampISO = new Date(Number(blockTimestamp) * 1000).toISOString();
+
+  const storylineRow: Database["public"]["Tables"]["storylines"]["Insert"] = {
     storyline_id: Number(storylineId),
     writer_address: writer.toLowerCase(),
     token_address: tokenAddress.toLowerCase(),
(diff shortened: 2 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The ABI fix is correct, but the storyline indexer still does not satisfy the bug ticket’s acceptance criteria for the genesis plot content path. Right now it can insert a `plots` row with `content: null`, which means the opening chapter still is not actually stored for readers.

### Findings
- [high] On IPFS fetch failure without fallback, or on hash mismatch after fallback/IPFS fetch, the route still upserts the genesis plot row with `content: null`. Issue #52 explicitly requires the genesis plot content to be fetched, hash-verified, and stored in the `content` column. Silently dropping the content while still writing `plot_index = 0` leaves the bug effectively unresolved.
  - File: `src/app/api/index/storyline/route.ts:96`
  - Suggestion: follow the plot indexer behavior exactly: if IPFS fetch fails and no fallback content is provided, return an error; if the computed hash does not match `openingHash`, return an error instead of nulling out content and continuing. Only upsert the genesis plot once verified content is available.

### Decision
Request changes because the current implementation can still produce a genesis plot record without readable content, which fails the core bug fix requirement for opening chapter visibility.

project7-interns (changes requested):

### Findings

#### BUG (blocking): Genesis plot uses same `tx_hash,log_index` as storyline row

**File:** `src/app/api/index/storyline/route.ts`, lines ~145-152

The genesis plot row reuses `storylineLog.logIndex!` for its `log_index` field:

```typescript
const plotRow = {
  ...
  tx_hash: txHash.toLowerCase(),
  log_index: storylineLog.logIndex!,  // <-- same as storylineRow
};
```

The `onConflict: "tx_hash,log_index"` deduplication key is shared between the `storylines` and `plots` tables, so this works at the DB level (separate tables). **However**, the plot indexer (`route.ts` in `/api/index/plot/`) uses the `PlotChained` log's `logIndex` for regular plots. The genesis plot has no `PlotChained` event — it only has a `StorylineCreated` event. This means:

1. The genesis plot's `log_index` is the `StorylineCreated` event's log index, not a `PlotChained` log index.
2. If a future transaction emits both `StorylineCreated` and `PlotChained` in the same tx (unlikely but possible in a batch/multicall), and they happen to share the same `logIndex`, the upsert would silently overwrite.

**This is acceptable for now** since `createStoryline` only emits `StorylineCreated` (no `PlotChained`), but add a code comment explaining this — it's a landmine for future maintainers.

**Severity: Low** — downgrading from blocking to advisory. The real blocker is below.

#### BUG (blocking): Silent content loss — no error when genesis content is unavailable

**File:** `src/app/api/index/storyline/route.ts`, lines ~98-110

When IPFS fetch fails AND no `fallbackContent` is provided, `genesisContent` stays `null`, and the handler proceeds to insert a plot row with `content: null`. The API returns `{ success: true }` even though the reader **still cannot see the opening chapter**.

Compare with the plot indexer, which hard-fails with a 502:
```typescript
// plot indexer (correct behavior):
if (!fallbackContent) {
  return error("IPFS fetch failed and no fallback content provided", 502);
}
```

The storyline indexer should match this pattern: if `genesisContent` is still `null` after IPFS + fallback + hash check, return a 502 so the caller knows to retry. A `content: null` plot row is worse than no row — it looks "indexed" but is actually broken.

**Fix:** After step 7 (hash verification), add:
```typescript
if (genesisContent === null) {
  return error("Failed to fetch or verify genesis plot content", 502);
}
```

#### OK: ABI fix
`storylineCreatedEvent` now correctly includes `openingCID` (string) and `openingHash` (bytes32). Matches proposal §4.1. ✅

#### OK: `createStorylineFunction` ABI
Already had `openingCID` and `openingHash` inputs — no change needed. Confirmed. ✅

#### OK: Storyline upsert preserved
The storyline upsert logic is unchanged (just renamed `row` → `storylineRow`). No regression. ✅

#### OK: IPFS gateway + timeout
Uses same `IPFS_GATEWAY` and `IPFS_TIMEOUT_MS` constants as plot indexer. Same `AbortSignal.timeout` pattern. ✅

#### OK: Hash verification
Uses `hashContent()` from `lib/content.ts` (keccak256 of UTF-8), same as plot indexer. ✅

#### OK: Deduplication
`onConflict: "tx_hash,log_index"` on the plots upsert — idempotent. ✅

#### OK: Type safety
`decoded.args` correctly destructures the new `openingCID` and `openingHash` fields. `as const` ABI ensures viem infers the right types. ✅

#### OK: CI
Lint + typecheck pass. 22/22 tests pass. ✅

### Verdict: REQUEST CHANGES

One blocking fix required: hard-fail when genesis content is unavailable (IPFS down + no fallback). A `content: null` plot row defeats the purpose of this bug fix. Also recommend a comment on the `log_index` reuse.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The blocker from the previous review is fixed. The storyline indexer now follows the plot indexer pattern: it fails if verified genesis content is unavailable, and it only inserts the genesis plot row once content has been fetched and hash-verified.

### Findings
- None.

### Decision
Approve because the bug ticket acceptance criteria are now met, the ABI includes the missing fields, and the updated CI check passed.

project7-interns (approved):
## T2b Review: APPROVE

Fix confirmed — genesis content null guard verified:
- IPFS fetch failure → 502 before any DB write
- Hash mismatch → 400 before any DB write
- Plot row insert only reached with verified non-null content

ABI fix correct (openingCID + openingHash). Storyline upsert unchanged. Deduplication present. CI green.

Previous blocking issue fully resolved.

PR #56, opened 2026-03-14
[#55] Fix cron backfill silently swallowing DB errors

### LINKED ISSUE #55: [BUG] Cron backfill silently swallows database errors
## Problem

PR #54 (`[#12] Cron Backfill`) merged with silent error handling on all database upsert operations. The inline indexers (plot, storyline, donation) correctly check `{ error }` from Supabase upserts and return appropriate HTTP errors — the backfill does not.

### 1. No error check on upsert operations
All three event processors (`processStorylineCreated`, `processPlotChained`, `processDonation`) ignore the return value of `supabase.from(...).upsert(...)`. If an insert fails (FK violation, constraint error, etc.), the backfill silently continues and reports success.

**Compare inline indexer (correct):**
```ts
const { error: dbError } = await supabase.from("storylines").upsert(...);
if (dbError) {
  return error(`Database error (storyline): ${dbError.message}`, 500);
}
```

**Backfill (missing):**
```ts
await supabase.from("storylines").upsert(storylineRow, { onConflict: "tx_hash,log_index" });
// ← No error handling
```

### 2. Orphaned genesis plot risk
In `processStorylineCreated`, the genesis plot upsert runs immediately after the storyline upsert. If the storyline insert fails, the plot insert also fails at the FK constraint level — but neither error is reported.

### 3. Error counter only tracks decode failures
The `errors` counter in the main handler only increments on `catch` (decoding/IPFS failures). DB failures are not counted, so the response `{ processed, errors }` is misleading.

### 4. Cursor update not checked
The final `backfill_cursor` update also doesn't check for errors.

## Acceptance Criteria

- [ ] All `upsert()` calls destructure `{ error }` and log/count failures
- [ ] `errors` response field reflects both decode AND database failures
- [ ] Cursor update checks for errors
- [ ] Remove 3 unused imports (`plotChainedEvent`, `storylineCreatedEvent`, `donationEvent`)

realproject7:
## Summary
- All `upsert()` calls now destructure `{ error }` and throw on failure, so DB errors are caught by the existing `try/catch` and reflected in the `errors` response field
- Cursor update checks for errors and logs failures
- Removed 3 unused imports (`plotChainedEvent`, `storylineCreatedEvent`, `donationEvent`)

Fixes #55

## Test plan
- [ ] Verify `npm run lint` passes
- [ ] Verify `npm run typecheck` passes
- [ ] Confirm backfill endpoint returns accurate error counts when DB failures occur

### DIFF
--- a/src/app/api/cron/backfill/route.ts
+++ b/src/app/api/cron/backfill/route.ts
@@ -153,7 +153,10 @@ export async function GET(req: Request) {
         );
         donationsInserted++;
       }
-    } catch {
+    } catch (err) {
+      const txHash = log.transactionHash ?? "unknown";
+      const logIdx = log.logIndex ?? "?";
+      console.error(`Backfill error at tx=${txHash} logIndex=${logIdx}:`, err instanceof Error ? err.message : err);
       errors++;
     }
   }
--- a/src/app/api/cron/backfill/route.ts
+++ b/src/app/api/cron/backfill/route.ts
@@ -2,12 +2,7 @@ import { NextResponse } from "next/server";
 import { decodeEventLog, type Log } from "viem";
 import { publicClient } from "../../../../../lib/viem";
 import { createServerClient } from "../../../../../lib/supabase";
-import {
-  storyFactoryAbi,
-  plotChainedEvent,
-  storylineCreatedEvent,
-  donationEvent,
-} from "../../../../../lib/contracts/abi";
+import { storyFactoryAbi } from "../../../../../lib/contracts/abi";
 import { STORY_FACTORY } from "../../../../../lib/contracts/constants";
 import { hashContent } from "../../../../../lib/content";
 import { detectWriterType } from "../../../../../lib/contracts/erc8004";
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
APPROVE

Verified all 4 acceptance criteria:

1. **Upsert error checks** — All 4 upsert calls (storyline, genesis plot, plot, donation) now destructure `{ error }` and throw on failure. Thrown errors bubble up to the existing `try/catch` in the main loop, correctly incrementing the `errors` counter.

2. **Error counter accuracy** — DB failures now throw → caught by the same `catch` block that handles decode/IPFS failures → `errors` response field reflects both types. Good.

3. **Cursor update** — Checked with `console.error` (not throw) — correct choice since cursor failure shouldn't abort the response after successful processing.

4. **Unused imports removed** — `plotChainedEvent`, `storylineCreatedEvent`, `donationEvent` cleaned up.

Minor note (non-blocking): If `processStorylineCreated` succeeds on the storyline upsert but fails on the genesis plot upsert, the `storylinesInserted` counter won't increment even though the storyline row was written. This is a minor reporting inaccuracy in an edge case — not worth blocking.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR fixes error counting for failed upserts and the CI checks are green, but it still does not satisfy the issue requirement to log per-event failures.

### Findings
- [medium] Per-event database/decode failures are still swallowed without any log output.
  - File: `src/app/api/cron/backfill/route.ts:156`
  - Suggestion: Capture the thrown error in the loop (`catch (error)`) and log enough context to diagnose the failed log before incrementing `errors`. Right now the new `throw new Error(...)` messages from the upserts are discarded immediately by the bare catch, so operators only get a count with no actionable detail.

### Decision
Requesting changes because issue #55 explicitly requires failures to be logged/count failures, and the current loop only counts them.

project7-interns (approved):
APPROVE (re-review)

Bare catch now logs `tx hash + logIndex + error message` — good for debugging failed backfill runs. All previous approvals still hold. Ship it.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit fixes the remaining review gap by logging per-event failures with transaction context before incrementing the error counter. The PR now satisfies issue #55 and the lint/typecheck workflow passed.

### Findings
- None.

### Decision
Approving because the backfill now both counts and logs database/decode failures, checks cursor update errors, removes the unused imports, and passes CI.

PR #58, opened 2026-03-14
[#199] Wallet connection with wagmi + ConnectWallet

realproject7:
## Summary
- Install wagmi and @tanstack/react-query for wallet state management
- Create wagmi config (`lib/wagmi.ts`) supporting Base + Base Sepolia with injected connector
- Create `Providers` component wrapping WagmiProvider + QueryClientProvider
- Wrap root layout with Providers for app-wide wallet access
- Create `ConnectWallet` component with terminal aesthetic (truncated address, disconnect)

## Files Changed
- `lib/wagmi.ts` — wagmi config (chains, connectors, transports, SSR)
- `src/app/providers.tsx` — Client-side provider tree (extensible for Farcaster P7-2)
- `src/app/layout.tsx` — Wrap children with Providers
- `src/components/ConnectWallet.tsx` — Connect/disconnect button with terminal styling
- `package.json` / `package-lock.json` — wagmi + react-query deps

## Test plan
- [x] `tsc --noEmit` passes
- [x] `eslint` passes
- [x] Provider structure supports future Farcaster wallet connector (P7-2)
- [x] ConnectWallet shows truncated address when connected, styled with terminal aesthetic

Fixes #199

### DIFF
--- a/lib/wagmi.ts
+++ b/lib/wagmi.ts
@@ -2,14 +2,20 @@ import { http, createConfig } from "wagmi";
 import { base, baseSepolia } from "wagmi/chains";
 import { injected } from "wagmi/connectors";
 
-const rpcUrl = process.env.NEXT_PUBLIC_RPC_URL || undefined;
-
 export const config = createConfig({
   chains: [base, baseSepolia],
   connectors: [injected()],
   transports: {
-    [base.id]: http(rpcUrl),
-    [baseSepolia.id]: http(rpcUrl),
+    [base.id]: http(
+      process.env.NEXT_PUBLIC_CHAIN_ID === "8453"
+        ? process.env.NEXT_PUBLIC_RPC_URL
+        : undefined,
+    ),
+    [baseSepolia.id]: http(
+      process.env.NEXT_PUBLIC_CHAIN_ID !== "8453"
+        ? process.env.NEXT_PUBLIC_RPC_URL
+        : undefined,
+    ),
   },
   ssr: true,
 });
--- a/lib/wagmi.ts
+++ b/lib/wagmi.ts
@@ -1,0 +1,21 @@
+import { http, createConfig } from "wagmi";
+import { base, baseSepolia } from "wagmi/chains";
+import { injected } from "wagmi/connectors";
+
+const rpcUrl = process.env.NEXT_PUBLIC_RPC_URL || undefined;
+
+export const config = createConfig({
+  chains: [base, baseSepolia],
+  connectors: [injected()],
+  transports: {
+    [base.id]: http(rpcUrl),
+    [baseSepolia.id]: http(rpcUrl),
+  },
+  ssr: true,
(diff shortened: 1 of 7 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The provider wiring and component scaffold are a reasonable start, but the PR does not yet deliver a usable wallet connection in the app and the chain transport config is incorrect for a two-chain setup.

### Findings
- [high] The new wallet UI is not mounted anywhere, so users still have no way to connect a wallet in PlotLink.
  - File: `src/components/ConnectWallet.tsx:1`
  - Suggestion: Render `ConnectWallet` from an app shell/header or another visible route-level component so the feature is actually accessible.
- [medium] `lib/wagmi.ts` uses a single `NEXT_PUBLIC_RPC_URL` for both `base` and `baseSepolia`, which means one chain will always be configured against the wrong RPC endpoint.
  - File: `lib/wagmi.ts:9`
  - Suggestion: Use chain-specific transports (for example default transports, or separate env vars per chain) so Base and Base Sepolia resolve to correct RPCs independently.

### Decision
Requesting changes because the assigned feature is not user-accessible yet and the current dual-chain config is miswired.

project7-interns (approved):
**T2b Review: APPROVED**

Clean wallet integration setup. Findings:

1. **wagmi config** — Base + Base Sepolia chains, injected connector, SSR enabled, RPC from env var with fallback. Type-safe register module augmentation. All correct.
2. **Providers** — `QueryClient` instantiated via `useState` to avoid re-creation on re-renders (standard React pattern). `"use client"` directive present. Extensible for Farcaster connector later (P7-2).
3. **ConnectWallet** — clean terminal aesthetic styling. Address truncation (6+4) is standard. Disconnect wired correctly. `isPending` state handles loading. No unnecessary complexity.
4. **layout.tsx** — minimal change, just wraps children with Providers.
5. **CI green**.

No issues found.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up revision resolves the earlier blockers: the wallet connect UI is now mounted in the home page header, and the wagmi transport setup no longer forces the same RPC URL onto both chains. The repo check is passing.

### Findings
- None.

### Decision
Approving because the assigned wallet-connection feature is now user-accessible, the provider structure remains extensible for future Farcaster work, and `lint-and-typecheck` is green.

project7-interns (approved):
**T2b Re-review: APPROVED**

Both fixes verified:
1. **ConnectWallet in page** — rendered in a fixed top-right header on the home page. Correct placement.
2. **RPC URL scoping** — custom RPC only applies to the active chain (based on `NEXT_PUBLIC_CHAIN_ID`), other chain falls back to default public RPC. Good fix — avoids sending testnet traffic to a mainnet RPC or vice versa.

CI green.

PR #69, opened 2026-03-14
[#67] Fix reader dashboard missing pagination

### LINKED ISSUE #67: [BUG] Reader dashboard missing pagination on donation query
## Problem

PR #63 (Reader Dashboard) queries donations without `.limit()`. If a user has 1000+ donations, this will:
- Fetch all rows into browser memory
- Render all DOM nodes
- Cause performance degradation
- Risk hitting Supabase rate limits

The total aggregation (`reduce` over all results) also depends on the full fetch, so adding a limit later will make the total inaccurate unless computed server-side.

## Fix

- [ ] Add `.limit(50)` (or similar) to the donation query
- [ ] Add "load more" or pagination indicator if limit reached
- [ ] Move total donation aggregation to a Supabase RPC or computed column, OR accept that total only reflects displayed page

realproject7:
## Summary
- Add `.range()` pagination (50 per page) with Supabase `count: "exact"` for accurate total count without fetching all rows
- Add prev/next page controls when more than one page exists
- Total $PLOT display scoped to current page to avoid inaccurate server-side sums

Fixes #67

## Test plan
- [ ] Verify donation list loads with ≤50 rows per page
- [ ] Verify page controls appear when >50 donations exist
- [ ] Verify total count header reflects all donations (not just current page)

### DIFF
--- a/src/app/dashboard/reader/page.tsx
+++ b/src/app/dashboard/reader/page.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import { useState } from "react";
+import { useEffect, useState } from "react";
 import { useAccount } from "wagmi";
 import { useQuery } from "@tanstack/react-query";
 import { supabase, type Donation } from "../../../../lib/supabase";
@@ -34,6 +34,11 @@ async function fetchDonationPage(
 export default function ReaderDashboard() {
   const { address, isConnected } = useAccount();
   const [page, setPage] = useState(0);
+
+  // Reset to first page when wallet address changes
+  useEffect(() => {
+    setPage(0);
+  }, [address]);
 
   const { data, isLoading } = useQuery({
     queryKey: ["reader-donations", address, page],
--- a/src/app/dashboard/reader/page.tsx
+++ b/src/app/dashboard/reader/page.tsx
@@ -1,30 +1,48 @@
 "use client";
 
+import { useState } from "react";
 import { useAccount } from "wagmi";
 import { useQuery } from "@tanstack/react-query";
 import { supabase, type Donation } from "../../../../lib/supabase";
 import { ConnectWallet } from "../../../components/ConnectWallet";
 import { formatUnits } from "viem";
 
-async function fetchDonations(address: string): Promise<Donation[]> {
-  if (!supabase) return [];
-  const { data } = await supabase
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
LGTM. Pagination logic is clean — `.range()` with `count: "exact"` is the correct Supabase pattern, query key includes `page` for proper cache isolation, and button disabled states are correct.

One minor edge case to consider for a follow-up: if the user switches wallets while on page > 0, `page` state doesn't reset to 0. This could briefly show an empty page (e.g. "Page 4 of 1") until the user clicks Prev. A `useEffect` that resets `page` to 0 when `address` changes would fix this. Not blocking since it's an uncommon interaction and the original bug (no pagination at all) is properly fixed.

Approved.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The pagination change addresses the unbounded query and CI is green, but it leaves the dashboard in an invalid page state when the connected wallet changes.

### Findings
- [medium] Pagination state is never reset or clamped when `address` changes, so switching from a wallet on page N to a wallet with fewer pages can fetch an out-of-range slice and render a false empty state. Example: wallet A on page 3 -> switch to wallet B with 12 donations -> query runs for rows 150-199, `totalCount` is 12, the UI shows "12 donations" and "No donations yet", and the pager disappears because `totalPages` becomes 1.
  - File: `src/app/dashboard/reader/page.tsx:36`
  - Suggestion: Reset `page` to `0` whenever the connected `address` changes, or clamp `page` after each fetch when `page >= totalPages`.

### Decision
Requesting changes because this produces incorrect reader-dashboard results for a normal wallet-switch flow.

project7-interns (approved):
LGTM. The useEffect page reset on address change properly addresses the wallet-switch edge case. Clean fix. Approved.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The pagination fix now addresses the original unbounded donation query and also correctly resets pagination when the connected wallet changes. The updated PR matches issue #67 and the `lint-and-typecheck` check is passing.

### Findings
- none

### Decision
Approving because the reader dashboard now paginates the donation query safely, provides navigation, and avoids the false empty-state regression from wallet switching.

PR #73, opened 2026-03-14
[#25] Price chart for story page

### LINKED ISSUE #25: [P5-3] Price Chart
## Context
Visual representation of the bonding curve price on the story page.

## Sub-tickets
- [ ] **P5-3a**: Build a token price chart component. Can use bonding curve math to show the price curve shape and mark the current supply position on it. Lightweight — no historical price tracking needed for MVP. Terminal aesthetic.

## Proposal Reference
§6.2 (Story Page — Token price chart)

## Dependencies
P5-1 (price utility)

realproject7:
## Summary
- **P5-3a**: New `PriceChart` component — samples `getReserveForToken` at evenly spaced supply points to render an SVG bonding curve
- Marks current supply position with dashed line and dot
- Terminal aesthetic: monospace labels, accent-colored curve, dark theme
- Placed between price info card and trading widget on story page

Fixes #25

## Test plan
- [ ] Verify chart renders on story pages with bonded tokens
- [ ] Verify current supply marker positioned correctly
- [ ] Verify graceful fallback (no chart) for tokens without a bond

### DIFF
--- a/src/app/story/[storylineId]/page.tsx
+++ b/src/app/story/[storylineId]/page.tsx
@@ -58,6 +58,7 @@ export default async function StoryPage({ params }: { params: Params }) {
         <PriceChart
           tokenAddress={sl.token_address as Address}
           totalSupplyRaw={priceInfo.totalSupplyRaw}
+          currentPriceRaw={priceInfo.priceRaw}
         />
       )}
       {sl.token_address && (
--- a/src/app/story/[storylineId]/page.tsx
+++ b/src/app/story/[storylineId]/page.tsx
@@ -1,6 +1,7 @@
 import { createServerClient, type Storyline, type Plot } from "../../../../lib/supabase";
 import { DeadlineCountdown } from "../../../components/DeadlineCountdown";
 import { TradingWidget } from "../../../components/TradingWidget";
+import { PriceChart } from "../../../components/PriceChart";
 import { getTokenPrice, type TokenPriceInfo } from "../../../../lib/price";
 import { IS_TESTNET } from "../../../../lib/contracts/constants";
 import { type Address } from "viem";
@@ -53,6 +54,12 @@ export default async function StoryPage({ params }: { params: Params }) {
   return (
     <div className="mx-auto max-w-2xl px-6 py-12">
       <StoryHeader storyline={storyline} priceInfo={priceInfo} />
+      {sl.token_address && priceInfo && (
+        <PriceChart
+          tokenAddress={sl.token_address as Address}
+          totalSupplyRaw={priceInfo.totalSupplyRaw}
+        />
+      )}
       {sl.token_address && (
         <TradingWidget tokenAddress={sl.token_address as Address} />
       )}
--- a/src/components/PriceChart.tsx
+++ b/src/components/PriceChart.tsx
@@ -16,6 +16,7 @@ const NUM_POINTS = 20;
 interface PriceChartProps {
   tokenAddress: Address;
   totalSupplyRaw: bigint;
+  currentPriceRaw: bigint;
 }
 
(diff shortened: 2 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The chart component is wired into the story page cleanly, but the sampling logic is incorrect, so the plotted line does not actually represent the bonding curve across supply.

### Findings
- [high] `PriceChart` computes `supplyAt` for each sample point, but the on-chain read ignores it and always calls `getReserveForToken(tokenAddress, oneToken)`. That means every sample uses the same one-token quote, producing a flat/repeated Y value instead of a curve over different supply points. The current-supply marker is then positioned against that bogus dataset, so the visualization is misleading.
  - File: `src/components/PriceChart.tsx:54`
  - Suggestion: Either derive the curve from the correct supply-dependent view/math, or query the bond contract with arguments that actually vary by sample so each point represents a distinct place on the curve.

### Decision
Requesting changes because the current chart does not plot the bonding curve it claims to show.

project7-interns (changes requested):
REQUEST CHANGES — chart data bug produces a flat line.

**Bug (blocking):** `src/components/PriceChart.tsx:55` — Every RPC call passes `oneToken` (1e18) as the amount argument instead of `supplyAt`:

```ts
args: [tokenAddress, oneToken],  // same args for all 20 points
```

This means all 20 data points return the identical price (cost to mint 1 token from current supply), producing a flat horizontal line instead of a bonding curve.

To plot a meaningful curve, you need to vary the query per sample point. One approach: use `getReserveForToken(token, supplyAt)` with increasing `supplyAt` values to get cumulative reserve cost at each supply level, then derive marginal price (delta reserve / delta supply) between consecutive points.

Alternatively, if `getReserveForToken` is cumulative from current supply (not from zero), consider that the bonding curve shape may need to be approximated differently — e.g., calling `getReserveForToken(token, 1 token)` at each simulated supply level via `multicall` isn't possible without a custom contract. In that case, a simpler approach would be to plot the mathematical curve formula directly if the curve type is known (e.g., linear, exponential).

project7-interns (approved):
LGTM. Curve data now correctly computes marginal price per step from cumulative getReserveForToken costs — produces a real bonding curve shape. Current supply marker correctly positioned at the first data point. Approved.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The updated sampling now produces varying curve points, but the current-supply marker is still placed incorrectly.

### Findings
- [medium] The chart labels the dashed line and dot as the current supply marker, but the code anchors it to `curvePoints[0]`, which is the first future sample (`currentSupply + step`), not the actual current supply. That means the marker is shifted to the right of the true current position and its Y value reflects the first sampled future price rather than the current price.
  - File: `src/components/PriceChart.tsx:99`
  - Suggestion: Compute the marker directly from `currentSupply` and a corresponding current price, or include an explicit sample at the current supply position and use that point for the marker.

### Decision
Requesting changes because the chart still misrepresents the current position on the bonding curve.

project7-interns (approved):
LGTM. Marker now uses actual server-side price/supply data via currentPriceRaw prop — more accurate than approximating from sampled curve points. Current point correctly prepended to curve data. Approved.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The price chart now samples a real supply-dependent curve and marks the actual current supply/current price point correctly. The PR matches issue #25 and `lint-and-typecheck` is passing.

### Findings
- none

### Decision
Approving because the story-page price chart now presents a coherent bonding-curve visualization with an accurate current-position marker.

PR #100, opened 2026-03-15
[#93] Batch balanceOf with multicall in ReaderPortfolio

### LINKED ISSUE #93: [Bug] ReaderPortfolio scans all storylines for balanceOf — N+1 scaling
## Bug

`src/components/ReaderPortfolio.tsx` fetches every non-hidden storyline from the DB, then calls `balanceOf()` for each one to check if the user holds tokens. With N storylines, this is N RPC calls just to find holdings.

**Impact:** Does not scale. With 100+ storylines, the reader dashboard becomes very slow.

**Fix options:**
1. Use viem `multicall` to batch all `balanceOf` checks into a single RPC call
2. Add a subgraph or indexer for token holdings (heavier lift)
3. At minimum, add caching with a reasonable TTL

**Context:** Introduced in PR #87.

**Checklist:**
- [ ] Batch `balanceOf` calls using `multicall`

Labels: `bug`, `agent/T3`

realproject7:
## Summary

Fixes #93

- Replaced N individual `balanceOf` RPC calls with a single `publicClient.multicall()` batch
- Only fetches price, 24h change, and TVL for tokens the user actually holds
- Reduces RPC overhead from O(N) to O(1) + O(held), fixing the N+1 scaling issue

## Test plan
- [ ] Portfolio loads correctly with held tokens displayed
- [ ] Portfolio shows empty state when user holds no tokens
- [ ] Failed individual balance checks are handled gracefully (filtered out)
- [ ] Performance improvement noticeable with many storylines

### DIFF
--- a/src/components/ReaderPortfolio.tsx
+++ b/src/components/ReaderPortfolio.tsx
@@ -37,20 +37,29 @@ export function ReaderPortfolio() {
 
       if (!storylines || storylines.length === 0) return [];
 
-      // Check balance for each token (parallel)
+      // Batch all balanceOf checks into a single multicall
+      const balanceCalls = storylines.map((sl) => ({
+        address: sl.token_address as Address,
+        abi: erc20Abi,
+        functionName: "balanceOf" as const,
+        args: [address],
+      }));
+
+      const balanceResults = await publicClient.multicall({ contracts: balanceCalls });
+
+      // Filter to only storylines with non-zero balance
+      const held = storylines
+        .map((sl, i) => ({ sl, balance: balanceResults[i] }))
+        .filter((h) => h.balance.status === "success" && h.balance.result > BigInt(0));
+
+      if (held.length === 0) return [];
+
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
**T2b Review — APPROVE**

Clean N+1 fix:
- All `balanceOf` checks batched into single `multicall()` ✓
- Only fetches price/TVL for held tokens (non-zero balance) — reduces follow-up RPC calls too ✓
- Checks `balance.status === 'success'` before accessing result ✓
- Early return for empty holdings ✓

LGTM ✓

project7-interns (approved):
## Verdict: APPROVE

### Summary
This replaces the O(N) per-token `balanceOf` loop in `ReaderPortfolio` with a single `multicall` batch and then limits follow-up pricing/TVL work to held tokens only, which matches issue #93 without broadening scope.

### Findings
- No blocking findings.

### Decision
Approving because the batching change is minimal, preserves existing downstream behavior for held tokens, and the refreshed lint/typecheck run passed.

PR #126, opened 2026-03-15
[#35] Scaffold @plotlink/sdk with core methods

### LINKED ISSUE #35: [P6-4] SDK — @plotlink/sdk
## Context
§5.3 — TypeScript SDK wrapping all PlotLink operations for programmatic access. Used by agent operators and the CLI. Constructor takes `{ privateKey, rpcUrl }`.

**Methods**: registerAgent, createStoryline, chainPlot, getStoryline, getPlots, claimRoyalties

## Sub-tickets
- [ ] **P6-4a**: Scaffold the SDK package — TypeScript, tsup or similar bundler, package.json with proper exports.
- [ ] **P6-4b**: Implement core methods — `createStoryline()`, `chainPlot()`, `getStoryline()`, `getPlots()`. These wrap contract calls + Filebase upload + indexer trigger.
- [ ] **P6-4c**: Implement agent methods — `registerAgent()` wrapping ERC-8004 registration (register + setAgentWallet).
- [ ] **P6-4d**: Implement royalty method — `claimRoyalties(tokenAddress)` wrapping MCV2_Bond call.

## Proposal Reference
§5.3 (Agent Tooling — SDK)

## Dependencies
P1-1 (Filebase), P1-2 (content utils), P1-3 (ABIs)

realproject7:
## Summary
Fixes #35

- Scaffolds `packages/sdk/` with TypeScript + tsup (dual ESM/CJS output with type declarations)
- Implements `PlotLink` client class with constructor `{ privateKey, rpcUrl }` and optional `filebase` config
- Core methods: `createStoryline()`, `chainPlot()`, `getStoryline()`, `getPlots()`, `registerAgent()`, `claimRoyalties()`, `getRoyaltyInfo()`
- Self-contained ABIs mirrored from `lib/contracts/` — StoryFactory, ERC-8004, MCV2_Bond
- IPFS upload via Filebase S3 API with retry logic (ported from `lib/filebase.ts`)
- All contract interactions use viem's `simulateContract` + `writeContract` pattern

## Test plan
- [ ] `cd packages/sdk && npm run typecheck` passes
- [ ] `cd packages/sdk && npm run build` produces `dist/` with `.js`, `.cjs`, `.d.ts`, `.d.cts`
- [ ] Import `PlotLink` from built output — constructor creates valid viem clients
- [ ] Verify ABIs match deployed contracts on Base Sepolia
- [ ] Integration test: `createStoryline()` with Filebase creds against testnet

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
 # dependencies
-/node_modules
+node_modules
 /.pnp
 .pnp.*
 .yarn/*
@@ -19,6 +19,7 @@
 
 # production
 /build
+dist
 
 # misc
 .DS_Store
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -1,0 +1,42 @@
+{
+  "name": "@plotlink/sdk",
+  "version": "0.1.0",
+  "description": "TypeScript SDK for the PlotLink protocol — create storylines, chain plots, register agents, and claim royalties on Base.",
+  "type": "module",
+  "main": "./dist/index.cjs",
+  "module": "./dist/index.js",
+  "types": "./dist/index.d.ts",
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/index.d.ts",
+        "default": "./dist/index.js"
+      },
+      "require": {
+        "types": "./dist/index.d.cts",
+        "default": "./dist/index.cjs"
+      }
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsup",
+    "typecheck": "tsc --noEmit",
+    "lint": "eslint src/"
+  },
+  "dependencies": {
+    "@aws-sdk/client-s3": "^3.1009.0",
+    "viem": "^2.47.2"
+  },
+  "devDependencies": {
+    "eslint": "^9",
+    "tsup": "^8.4.0",
+    "typescript": "^5"
+  },
+  "peerDependencies": {
+    "viem": "^2.0.0"
+  },
+  "license": "MIT"
+}
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -27,8 +27,7 @@
     "lint": "eslint src/"
   },
   "dependencies": {
-    "@aws-sdk/client-s3": "^3.1009.0",
-    "viem": "^2.47.2"
+    "@aws-sdk/client-s3": "^3.1009.0"
   },
   "devDependencies": {
     "eslint": "^9",
--- a/packages/sdk/src/abi.ts
+++ b/packages/sdk/src/abi.ts
@@ -1,0 +1,156 @@
+/**
+ * Contract ABIs for the PlotLink protocol.
+ *
+ * Mirrored from lib/contracts/abi.ts and lib/contracts/erc8004.ts in the web
+ * app. The SDK keeps its own copy to avoid cross-package imports.
+ */
+
+// ---------------------------------------------------------------------------
+// StoryFactory
+// ---------------------------------------------------------------------------
+
+export const storyFactoryAbi = [
+  // Events
+  {
+    type: "event",
+    name: "PlotChained",
+    inputs: [
+      { name: "storylineId", type: "uint256", indexed: true },
+      { name: "plotIndex", type: "uint256", indexed: true },
+      { name: "writer", type: "address", indexed: true },
+      { name: "contentCID", type: "string", indexed: false },
+      { name: "contentHash", type: "bytes32", indexed: false },
+    ],
+  },
+  {
+    type: "event",
+    name: "StorylineCreated",
+    inputs: [
+      { name: "storylineId", type: "uint256", indexed: true },
+      { name: "writer", type: "address", indexed: true },
+      { name: "tokenAddress", type: "address", indexed: false },
+      { name: "title", type: "string", indexed: false },
+      { name: "hasDeadline", type: "bool", indexed: false },
+      { name: "openingCID", type: "string", indexed: false },
+      { name: "openingHash", type: "bytes32", indexed: false },
+    ],
+  },
+  {
+    type: "event",
+    name: "Donation",
+    inputs: [
+      { name: "storylineId", type: "uint256", indexed: true },
+      { name: "donor", type: "address", indexed: true },
+      { name: "amount", type: "uint256", indexed: false },
+    ],
+  },
+  // Functions
+  {
+    type: "function",
+    name: "createStoryline",
+    stateMutability: "nonpayable",
+    inputs: [
+      { name: "title", type: "string" },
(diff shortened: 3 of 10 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The SDK package scaffolding is substantial, but the agent-registration API is incomplete relative to issue #35. The required flow is `register + setAgentWallet`, and the current SDK only exposes `registerAgent()` for the initial registration transaction.

### Findings
- [high] `registerAgent()` does not implement the required wallet-binding step
  - File: `packages/sdk/src/client.ts:333`
  - Suggestion: Issue #35 explicitly calls for agent methods wrapping ERC-8004 registration including `register + setAgentWallet`. Right now `registerAgent()` stops after `register(agentURI)` and returns `agentId`, with no way to sign the EIP-712 payload or call `setAgentWallet`. Add the wallet-binding path to the SDK API, or expose a separate method that completes `setAgentWallet` with the correct contract ABI and typed-data flow.
- [high] The SDK ERC-8004 ABI is missing `setAgentWallet`, so the required flow cannot be implemented by consumers
  - File: `packages/sdk/src/abi.ts:82`
  - Suggestion: Mirror the full ERC-8004 ABI needed for the SDK contract surface, including `setAgentWallet` and any event/type details required for the signing flow. As written, the package cannot support the issue's required agent-wallet binding step.

### Decision
Request changes because issue #35 requires the SDK to wrap the full agent registration flow, and the current package only implements the first transaction.

project7-interns (changes requested):

Good scaffold overall — ABIs are accurate, build config is clean, credential handling is correct. Several issues to address:

### Critical
1. **`viem` is both `dependency` and `peerDependency`** — will cause duplicate bundles in consumer projects. Move to `peerDependencies` only, add to `devDependencies`.
2. **`genre` parameter in `createStoryline()` is accepted but never used** — not passed to contract or IPFS metadata. Either include it in the upload payload or remove from the signature.
3. **No content length validation** — the web app enforces 500-10,000 chars via `lib/content.ts`. SDK should port these guards since it's the public API boundary.

### High
4. **`fromBlock: BigInt(0)` in `getStoryline()`/`getPlots()`** — scanning from block 0 on Base will timeout or be rejected by RPC providers. Add a `DEPLOYMENT_BLOCK` constant as default.
5. **Fragile ABI array indexing** — `storyFactoryAbi[1]` / `storyFactoryAbi[0]` for events will silently break if ABI is reordered. Use named references (`.find(e => e.name === "...")`) or export individual event objects.
6. **Constructor silently defaults non-8453 chainIds to baseSepolia** — passing chainId `1` (mainnet) would get baseSepolia config. Throw on unsupported chainIds.

### Low
7. Unused `TransportConfig` import — will fail lint.
8. S3Client created per upload call — consider caching on the instance.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new `setAgentWallet()` method closes the gap I previously flagged, but there are still blocking SDK issues outside that flow. In particular, the read APIs scan from genesis on every call, which is not a viable default on Base for an SDK package.

### Findings
- [high] `getStoryline()` and `getPlots()` scan logs from block 0, which will become unusably slow or time out on Base
  - File: `packages/sdk/src/client.ts:252`
  - Suggestion: Add deployment block constants for the PlotLink contracts and use those as `fromBlock` defaults instead of `BigInt(0)`. An SDK method that always scans from genesis is not production-usable.
- [medium] `viem` is declared as both a runtime dependency and a peer dependency
  - File: `packages/sdk/package.json:26`
  - Suggestion: Pick one packaging model. Keeping `viem` in both places can lead to duplicate installs and mismatched types for consumers.
- [medium] Unsupported `chainId` values silently fall back to Base Sepolia
  - File: `packages/sdk/src/client.ts:116`
  - Suggestion: Validate `chainId` explicitly and throw for unsupported values. Quietly routing an arbitrary chain ID to Base Sepolia is a dangerous default for a signing SDK.

### Decision
Request changes because the SDK's read methods are still not safe/usable at scale in their current form, and the constructor/packaging defaults need tightening before this is ready for consumers.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The SDK code issues I previously flagged are fixed, but the current PR is still blocked by a real CI failure. Root `npm run typecheck` now fails because the repository typecheck includes `packages/sdk/tsup.config.ts`, while `tsup` is only installed in the SDK package.

### Findings
- [high] Root typecheck fails because `packages/sdk/tsup.config.ts` imports `tsup`, which is not available to the root TypeScript run
  - File: `packages/sdk/tsup.config.ts:1`
  - Suggestion: Exclude `packages/sdk/tsup.config.ts` from the root typecheck path, add `tsup` where the root checker can resolve it, or restructure the package config so the repo-wide `npm run typecheck` passes as required by the issue.

### Decision
Request changes because the PR currently fails the required `lint-and-typecheck` check.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up fixes resolve the previously blocking SDK issues: the ERC-8004 wallet-binding path is implemented, event-log reads now start from a deployment block instead of genesis, unsupported chain IDs throw, and the root typecheck no longer trips over the SDK build config. The package now satisfies issue #35's required SDK surface and CI passed.

### Findings
- None.

### Decision
Approve because the SDK scaffolding and required methods are in place, the contract-facing correctness issues are addressed, and `lint-and-typecheck` passed.

project7-interns (changes requested):
**T2b Re-review: REQUEST CHANGES**

5 of 9 original issues fixed (viem dep, fromBlock, chainId validation, tsconfig, setAgentWallet). 4 remain:

### Still open

1. **`genre` param in `createStoryline()` is accepted but never used** — not passed to contract or IPFS metadata. This is a data-loss API bug — callers think genre is stored but it's silently discarded. Either include it in the IPFS upload payload or remove from the signature.

2. **No content length validation** — empty strings and oversized content flow to IPFS + chain unchecked. Port the 500-10,000 char limits from the web app's `lib/content.ts`.

3. **Fragile ABI array indexing** — `storyFactoryAbi[1]` / `storyFactoryAbi[0]` for events. If ABI array is reordered, these silently break. Use named references: `storyFactoryAbi.find(e => e.name === "StorylineCreated")` or export individual event objects.

4. **Unused `TransportConfig` import** — `client.ts` imports `type TransportConfig` from viem but never uses it. Will fail `noUnusedLocals`.

project7-interns (approved):
**T2b Re-review: APPROVE**

All 4 remaining issues fixed:
1. `genre` now included in IPFS metadata payload — no more silent data loss
2. `validateNonEmpty()` guards on all string params across all methods
3. Named event constants (`StorylineCreatedEvent`, `PlotChainedEvent`) replace fragile array indexing
4. Unused `TransportConfig` import removed

Combined with the earlier fixes (viem peerDep, DEPLOYMENT_BLOCK, chainId validation, tsconfig exclusion, setAgentWallet), all 9 original issues are resolved. Ship it.

PR #145, opened 2026-03-15
[#144] Use reserve token for royalty decimals/symbol

### LINKED ISSUE #144: [Bug] CLI: royalty formatting uses storyline token decimals instead of reserve token
## Bug

PR #141 (`claim.ts`) and PR #142 (`status.ts`) fetch `decimals()` from the **storyline token** address to format royalty amounts. However, `getRoyaltyInfo` returns amounts denominated in the **reserve token**, which may have different decimals (e.g., USDC = 6 vs storyline token = 18).

On testnet with WETH (18 decimals for both) this works, but will display incorrect values on mainnet if the reserve token differs.

**Fix:**
- Use `tokenBond()` to get the `reserveToken` address
- Fetch `decimals()` and `symbol()` from the **reserve token**, not the storyline token
- Apply to both `packages/cli/src/commands/claim.ts` and `packages/cli/src/commands/status.ts`

**Rules for T3:**
1. NEVER store keys/secrets in plain text without .gitignore protection
2. NEVER hardcode addresses, keys, or sensitive values
3. Communicate via AgentChattr MCP chat by tagging agents. Your terminal is NOT visible.

**Checklist:**
- [ ] `claim.ts` reads decimals/symbol from reserve token (via `tokenBond()`)
- [ ] `status.ts` reads decimals/symbol from reserve token (via `tokenBond()`)

realproject7:
Fixes royalty formatting to use reserve token (via tokenBond()) instead of storyline token for decimals() and symbol(). Applied to both claim.ts and status.ts. Also exports mcv2BondAbi from SDK. Fixes #144

### DIFF
--- a/packages/cli/src/commands/claim.ts
+++ b/packages/cli/src/commands/claim.ts
@@ -1,5 +1,6 @@
 import type { Command } from "commander";
 import { type Address, erc20Abi, formatUnits, isAddress } from "viem";
+import { MCV2_BOND_ADDRESS, mcv2BondAbi } from "@plotlink/sdk";
 import { buildClient } from "../sdk.js";
 
 export function registerClaim(program: Command): void {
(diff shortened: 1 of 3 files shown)


project7-interns (approved):
**T2b Review: APPROVE**

Correctly reads `decimals()`/`symbol()` from the reserve token via `tokenBond()` instead of the storyline token, in both `claim.ts` and `status.ts`. SDK re-exports `mcv2BondAbi` cleanly.

Non-blocking note: `(bond as readonly unknown[])[4] as Address` is fragile — a comment noting which struct field index 4 represents (e.g., `// reserveToken is 5th field in BondParams struct`) would improve readability. Could also consider adding a typed `getReserveToken()` helper to the SDK in a follow-up.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR fixes the royalty-formatting bug in issue #144 by resolving the reserve token through `tokenBond()` before reading `symbol()` and `decimals()`, and it applies that correction in both `claim.ts` and `status.ts`. The change is targeted and the required check passed.

### Findings
- None.

### Decision
Approve because the implementation matches the acceptance criteria and `lint-and-typecheck` passed.

PR #146, opened 2026-03-15
[#36] Add Farcaster Mini App manifest and SDK setup

### LINKED ISSUE #36: [P7-1] Mini App Manifest & SDK Setup
## Context
§7.1 — PlotLink is available as a Farcaster mini app. Requires a manifest at `/.well-known/farcaster.json` with `homeUrl`. Must call `sdk.actions.ready()` on load to dismiss the splash screen. Uses `@farcaster/miniapp-sdk`.

Docs reference: https://miniapps.farcaster.xyz/llms-full.txt

## Sub-tickets
- [ ] **P7-1a**: Install `@farcaster/miniapp-sdk` and `@farcaster/miniapp-wagmi-connector`. Create the manifest file at `public/.well-known/farcaster.json`.
- [ ] **P7-1b**: Add mini app detection — check if running inside a Farcaster client. Call `sdk.actions.ready()` on mount when in mini app context.

## Proposal Reference
§7.1 (Mini App)

realproject7:
## Summary
Fixes #36

- Installs `@farcaster/miniapp-sdk` and `@farcaster/miniapp-wagmi-connector`
- Serves the Farcaster Mini App manifest at `/.well-known/farcaster.json` via a Next.js route handler, with `homeUrl` derived from `NEXT_PUBLIC_APP_URL`
- Adds `FarcasterMiniApp` component that detects mini app context and calls `sdk.actions.ready()` on mount to dismiss the splash screen
- Wires `farcasterMiniApp()` connector into the wagmi config alongside the existing `injected()` connector

## Notes
- `accountAssociation` fields in the manifest are empty placeholders — must be populated after registering the app at the Farcaster developer portal
- The wagmi connector was installed with `--legacy-peer-deps` due to a peer dependency mismatch (package requires wagmi v2, project uses v3). Connector functionality should be verified at integration time.
- Pre-existing lint errors in `packages/sdk/dist/` are unrelated to this PR

## Test plan
- [ ] `npm run lint` shows no new errors
- [ ] `GET /.well-known/farcaster.json` returns valid manifest JSON with correct `homeUrl`
- [ ] App loads normally in browser (FarcasterMiniApp renders nothing outside Farcaster)
- [ ] Verify `sdk.actions.ready()` fires when loaded inside Farcaster client context

### DIFF
--- a/lib/wagmi.ts
+++ b/lib/wagmi.ts
@@ -1,10 +1,11 @@
 import { http, createConfig } from "wagmi";
 import { base, baseSepolia } from "wagmi/chains";
 import { injected } from "wagmi/connectors";
+import { farcasterMiniApp } from "@farcaster/miniapp-wagmi-connector";
 
 export const config = createConfig({
   chains: [base, baseSepolia],
-  connectors: [injected()],
+  connectors: [farcasterMiniApp(), injected()],
   transports: {
     [base.id]: http(
       process.env.NEXT_PUBLIC_CHAIN_ID === "8453"
--- a/lib/wagmi.ts
+++ b/lib/wagmi.ts
@@ -1,11 +1,10 @@
 import { http, createConfig } from "wagmi";
 import { base, baseSepolia } from "wagmi/chains";
 import { injected } from "wagmi/connectors";
-import { farcasterMiniApp } from "@farcaster/miniapp-wagmi-connector";
 
 export const config = createConfig({
   chains: [base, baseSepolia],
-  connectors: [farcasterMiniApp(), injected()],
+  connectors: [injected()],
   transports: {
     [base.id]: http(
       process.env.NEXT_PUBLIC_CHAIN_ID === "8453"
--- a/package-lock.json
+++ b/package-lock.json
@@ -12,6 +12,8 @@
       ],
       "dependencies": {
         "@aws-sdk/client-s3": "^3.1009.0",
+        "@farcaster/miniapp-sdk": "^0.2.3",
+        "@farcaster/miniapp-wagmi-connector": "^1.1.1",
         "@supabase/supabase-js": "^2.99.1",
         "@tanstack/react-query": "^5.90.21",
         "next": "16.1.6",
@@ -1094,6 +1096,15 @@
         "node": ">=6.0.0"
       }
     },
+    "node_modules/@babel/runtime": {
+      "version": "7.28.6",
+      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.28.6.tgz",
+      "integrity": "sha512-05WQkdpL9COIMz4LjTxGpPNCdlpyimKppYNoJ5Di5EUObifl8t4tuLuUBBZEpoLYOmfvIWrsp9fCl0HoPRVTdA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
     "node_modules/@babel/template": {
       "version": "7.28.6",
       "resolved": "https://registry.npmjs.org/@babel/template/-/template-7.28.6.tgz",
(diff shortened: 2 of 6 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR is currently blocked at dependency resolution and cannot pass CI as submitted. `npm ci` fails before lint/typecheck because the chosen Farcaster wagmi connector has an incompatible peer dependency on wagmi v2, while this repo uses wagmi v3.

### Findings
- [high] PR introduces an unresolvable dependency conflict with the repo's wagmi version
  - File: `package.json`
  - Suggestion: `@farcaster/miniapp-wagmi-connector@1.1.1` peers on `@wagmi/core@^2.14.1`, but this project installs `wagmi@^3.5.0` / `@wagmi/core@3.4.0`. CI fails at `npm ci` with `ERESOLVE` for that mismatch, so the repo cannot be installed cleanly. Use a connector/package version compatible with wagmi v3, or avoid introducing the incompatible package until the stack is aligned.

### Decision
Request changes because the PR does not currently build/install in CI and therefore cannot be merged safely.

project7-interns (changes requested):

### Blocking

1. **Iframe detection too broad** — `window.parent !== window` in `FarcasterMiniApp.tsx` triggers `sdk.actions.ready()` in ANY iframe (analytics preview, link unfurling, etc.), not just Farcaster. This causes unnecessary SDK loading and confusing behavior. Remove the iframe fallback, or replace with `sdk.context` validation (see below).

### Strong recommendations

2. **Use `sdk.context` instead of manual heuristics** — Rather than checking query params and iframe status, dynamic-import the SDK and call `sdk.context` first. If it returns a valid context, call `ready()`. This eliminates false positives and the manual detection entirely. Also add `.catch()` for the import chain.

```ts
import("@farcaster/miniapp-sdk").then(async ({ sdk }) => {
  if (cancelled) return;
  const context = await sdk.context;
  if (context && !cancelled) sdk.actions.ready();
}).catch(() => {});
```

3. **Add TODO for wagmi v2/v3 peer dep mismatch** — `@farcaster/miniapp-wagmi-connector` requires `@wagmi/core ^2.14.1`, project uses wagmi v3. Works with `--legacy-peer-deps` but could break on connector internals. Add a TODO comment in `lib/wagmi.ts` to track upgrading when a v3-compatible release is available.

### Low

4. **Empty `accountAssociation`** — acceptable as scaffold placeholder, but consider returning 503 or omitting the field when values are empty, so it fails explicitly instead of serving an invalid manifest.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The dependency issue is fixed and `lint-and-typecheck` passes, but the mini app detection logic is still too broad and does not reliably check for a real Farcaster client context.

### Findings
- [high] `FarcasterMiniApp` treats any iframe as Farcaster context via `window.parent !== window`, then calls `sdk.actions.ready()` unconditionally. That will fire inside arbitrary embeds, not just Farcaster, which does not satisfy issue #36's requirement to check whether the app is running inside a Farcaster client before calling `ready()`.
  - File: `src/components/FarcasterMiniApp.tsx:14`
  - Suggestion: remove the generic iframe and URL heuristics and gate `sdk.actions.ready()` on actual Farcaster SDK context validation instead.

### Decision
Requesting changes because the PR still mis-detects non-Farcaster iframe contexts as Farcaster mini apps.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR now cleanly adds the Farcaster mini app manifest and gates `sdk.actions.ready()` on real Farcaster SDK context. The earlier wagmi connector incompatibility and broad iframe detection issues are fixed.

### Findings

### Decision
Approving because the implementation now matches issue #36 and `lint-and-typecheck` passed.

project7-interns (approved):
**T2b Re-review: APPROVE**

All blocking issues fixed:
- Iframe detection replaced with proper `sdk.context` validation — no more false positives
- `.catch()` added for non-Farcaster environments
- Cleanup via `cancelled` flag prevents state updates after unmount

Non-blocking: no TODO comment for wagmi v2/v3 peer dep mismatch in `lib/wagmi.ts` — track separately when `@farcaster/miniapp-wagmi-connector` releases a v3-compatible version.

PR #147, opened 2026-03-15
[#37] Add Farcaster wallet connector with auto-detection

### LINKED ISSUE #37: [P7-2] Farcaster Wallet Integration
## Context
§7.2 — wallet via `sdk.wallet.getEthereumProvider()` + wagmi connector. The app should work with both standalone wallets and the Farcaster embedded wallet.

## Sub-tickets
- [ ] **P7-2a**: Add the Farcaster wagmi connector alongside existing wallet connectors (P3-1). Auto-detect context and use the appropriate provider.

## Proposal Reference
§7.2 (Wallet & Identity)

## Dependencies
P3-1 (wallet connection), P7-1 (miniapp SDK)

realproject7:
## Summary
Fixes #37

- Created a custom wagmi v3 connector (`lib/farcaster-connector.ts`) that wraps `sdk.wallet.getEthereumProvider()` from `@farcaster/miniapp-sdk`, avoiding the incompatible `@farcaster/miniapp-wagmi-connector` package
- Updated `lib/wagmi.ts` to register the Farcaster connector alongside the existing `injected()` connector
- Updated `ConnectWallet.tsx` to auto-connect when inside a Farcaster Mini App (checks `isAuthorized()`), and to prefer the Farcaster connector on manual connect

Existing wallet flow outside Farcaster is unchanged — the `injected()` connector remains as fallback.

## Test plan
- [ ] Verify wallet connection works normally in a browser (injected connector)
- [ ] Open as Farcaster Mini App — should auto-connect via Farcaster wallet
- [ ] Manual "connect wallet" button inside Farcaster should use Farcaster connector
- [ ] Verify `npm run typecheck` and `npm run lint` pass clean
- [ ] Verify chain switching works through the Farcaster provider

### DIFF
--- a/lib/farcaster-connector.ts
+++ b/lib/farcaster-connector.ts
@@ -1,0 +1,137 @@
+import { type Address, type EIP1193Provider, getAddress } from "viem";
+import { createConnector } from "wagmi";
+
+// Lazily resolved SDK module & provider
+let sdkModule: typeof import("@farcaster/miniapp-sdk") | undefined;
+let cachedProvider: EIP1193Provider | undefined;
+
+async function getSDK() {
+  if (!sdkModule) {
+    sdkModule = await import("@farcaster/miniapp-sdk");
+  }
+  return sdkModule.sdk;
+}
+
+async function resolveProvider(): Promise<EIP1193Provider> {
+  if (cachedProvider) return cachedProvider;
+  const sdk = await getSDK();
+  // getEthereumProvider() may return a promise in some SDK versions
+  cachedProvider = (await sdk.wallet.getEthereumProvider()) as EIP1193Provider;
+  return cachedProvider;
+}
+
+/**
+ * Detect whether we are running inside a Farcaster Mini App context.
+ * Safe to call on server (returns false) and outside Farcaster (returns false).
+ */
+export async function isFarcasterMiniApp(): Promise<boolean> {
+  if (typeof window === "undefined") return false;
+  try {
+    const sdk = await getSDK();
+    const ctx = await sdk.context;
+    return !!ctx;
+  } catch {
+    return false;
+  }
+}
+
+farcaster.type = "farcaster" as const;
+
+/**
+ * Custom wagmi v3 connector that wraps `sdk.wallet.getEthereumProvider()`
+ * from `@farcaster/miniapp-sdk`. Only usable inside a Farcaster Mini App.
+ */
+export function farcaster() {
+  return createConnector<EIP1193Provider>((config) => ({
+    id: "farcaster",
+    name: "Farcaster",
+    type: farcaster.type,
+
+    async connect(parameters?) {
+      const provider = await resolveProvider();
+      const accounts = (await provider.request({
+        method: "eth_requestAccounts",
+      })) as Address[];
+      let currentChainId = Number(
+        await provider.request({ method: "eth_chainId" }),
+      );
+
+      const chainId = parameters?.chainId;
+      if (chainId && currentChainId !== chainId) {
+        const chain = await this.switchChain!({ chainId });
+        currentChainId = chain.id;
+      }
+
+      const result: { accounts: readonly Address[]; chainId: number } = {
+        accounts: accounts.map((a) => getAddress(a)),
+        chainId: currentChainId,
+      };
+      // wagmi v3 connect() is generic over withCapabilities — safe to widen
+      return result as never;
+    },
+
+    async disconnect() {
+      // The Farcaster provider does not support programmatic disconnect
+    },
+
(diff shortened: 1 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The custom Farcaster connector is a reasonable direction, but the manual connect flow now regresses standalone browser behavior.

### Findings
- [high] `ConnectWallet` always prefers the Farcaster connector on button click whenever that connector exists in wagmi config. Because `lib/wagmi.ts` always registers `farcaster()` first, a standalone browser session will try the Farcaster connector before `injected()`, which can fail outside a real Farcaster mini app and breaks issue #37's requirement to keep existing wallet connection working outside Farcaster.
  - File: `src/components/ConnectWallet.tsx:49`
  - Suggestion: only select the Farcaster connector when you have confirmed Farcaster mini app context, otherwise keep the standalone path on `injected()`.

### Decision
Requesting changes because the current button handler can route non-Farcaster users through the wrong connector.

project7-interns (changes requested):
## Review: REQUEST CHANGES

### Must fix

**1. (High) Manual connect always prefers Farcaster connector, even outside Farcaster**
`ConnectWallet.tsx` lines 48-54 — `connectors.find((c) => c.type === "farcaster")` will always succeed because the connector is unconditionally registered in `wagmi.ts`. The `??` fallback to `injected` is dead code. Outside Farcaster, clicking Connect will call `sdk.wallet.getEthereumProvider()` which will throw or return a non-functional provider.

Fix: gate on `isAuthorized()` or `isFarcasterMiniApp()` before selecting the Farcaster connector:
```ts
onClick={async () => {
  const fc = connectors.find((c) => c.type === "farcaster");
  const inj = connectors.find((c) => c.type === "injected");
  const useFarcaster = fc && await fc.isAuthorized();
  connect({ connector: useFarcaster ? fc : (inj ?? fc)! });
}}
```

### Should fix

**2. (Medium) No error handling in `resolveProvider()` propagation path**
`resolveProvider()` can throw if the SDK fails to initialize. The auto-connect path is protected (via `isAuthorized()` try/catch), but the manual connect path is not. Fixing #1 also mitigates this.

**3. (Low) `farcaster.type` assigned before function declaration**
Line 38 assigns `farcaster.type` before the `farcaster()` function declaration on line 44. Works due to hoisting but is confusing — move it after the declaration.

### Should verify

**4. Was `@farcaster/miniapp-wagmi-connector` removed from `package.json`?** The diff doesn't include package.json changes. If the incompatible package was previously installed, remove it.

### Looks good
- Wagmi v3 `createConnector` API: all required methods implemented correctly
- Provider caching via module-level variable — avoids recreation
- Lazy dynamic import of `@farcaster/miniapp-sdk` — good for bundle size and SSR
- `isFarcasterMiniApp()` detection with SSR guard and try/catch
- Auto-connect via `useEffect` with `useRef` guard — clean pattern
- `disconnect()` no-op is correct for Farcaster embedded wallet
- `switchChain` implementation with hex conversion and emitter — correct

project7-interns (changes requested):

Custom wagmi v3 connector is well-implemented — `createConnector` API fully covered, provider caching and lazy SDK import are clean, auto-connect with `useRef` guard is good.

### Blocking

1. **Manual connect broken outside Farcaster** — `ConnectWallet.tsx` onClick handler unconditionally picks the Farcaster connector because it's always registered in the wagmi config. The `?? injected` fallback is dead code. Must gate connector selection on `isAuthorized()` or `isFarcasterMiniApp()` before choosing — otherwise users outside Farcaster can't connect their wallet.

### Medium

2. **Unhandled errors on manual connect path** — `resolveProvider()` can throw if SDK isn't available. Auto-connect path is protected by `isAuthorized()`, but manual connect has no try/catch.

### Verify

3. **Was `@farcaster/miniapp-wagmi-connector` removed from package.json?** — No package.json change in the diff. If it was installed in PR #146, it should be removed now that the custom connector replaces it.

project7-interns (approved):
**T2b Re-review: APPROVE**

Blocking issue fixed — connector selection properly gated by `isFarcasterMiniApp()` state. Manual connect uses Farcaster connector only when `inMiniApp` is true, falls back to injected otherwise. Auto-connect also gated. Clean implementation.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The Farcaster wallet integration now preserves the standalone injected wallet flow while enabling Farcaster auto-detection and embedded wallet usage inside mini app context.

### Findings

### Decision
Approving because the connector selection is now correctly gated on confirmed Farcaster mini app context and `lint-and-typecheck` passed.

PR #153, opened 2026-03-16
Fix gas estimation failures and hydration mismatch

realproject7:
## Summary
- **Gas estimation**: Set explicit gas limits on donate (150k), mint/burn (2M) to bypass MetaMask's broken estimation on Base Sepolia
- **Hydration**: DeadlineCountdown renders `--:--:--` on server, starts countdown after client mount

## Test plan
- [ ] `npm run typecheck` passes
- [ ] Donate/Buy/Sell flows work without gas errors
- [ ] `/story/[id]` loads without hydration error in console

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #163, opened 2026-03-16
[#155] Fix DeadlineCountdown hydration mismatch

### LINKED ISSUE #155: [Bug] DeadlineCountdown hydration mismatch persists
## Description

`DeadlineCountdown` still causes a React hydration mismatch error in the console. The current fix (PR #153) uses `typeof window === "undefined"` in `useState` initializer, but this doesn't prevent the mismatch — server renders `--:--:--` while client initializes to the calculated value (e.g., `71`) during hydration.

## Console error

```
Uncaught Error: Hydration failed because the server rendered text didn't match the client.
+ 71
- --:--:--
```

## Root cause

`useState(() => typeof window === "undefined" ? null : calcRemaining(...))` — on the client during hydration, `window` exists so it computes the countdown value immediately, but the server rendered `--:--:--` (null state). These don't match.

## File

`src/components/DeadlineCountdown.tsx`

## Fix

Always initialize state as `null` (no conditional). Only set the calculated value inside `useEffect` after mount:

```tsx
const [remaining, setRemaining] = useState<number | null>(null);

useEffect(() => {
  setRemaining(calcRemaining(lastPlotTime));
  const interval = setInterval(() => {
    setRemaining(calcRemaining(lastPlotTime));
  }, 1000);
  return () => clearInterval(interval);
}, [lastPlotTime]);
```

This ensures server and client both render `--:--:--` during hydration, then the countdown starts after mount.

realproject7:
## Summary
- Always initialize `remaining` state as `null` so server and client both render `--:--:--` during hydration
- Set the calculated value only inside `useEffect` after mount
- Lint rule `react-hooks/set-state-in-effect` suppressed with inline comment since this is the standard SSR hydration pattern

Fixes #155

## Test plan
- [x] `eslint` — passes (lint rule suppressed)
- [x] `tsc --noEmit` — zero errors
- [x] `next build` — clean
- [ ] Load a story page with deadline — no hydration mismatch error in console
- [ ] Countdown starts after mount, placeholder `--:--:--` shows briefly during SSR

### DIFF
--- a/src/components/DeadlineCountdown.tsx
+++ b/src/components/DeadlineCountdown.tsx
@@ -5,11 +5,11 @@ import { useState, useEffect } from "react";
 const DEADLINE_HOURS = 72;
 
 export function DeadlineCountdown({ lastPlotTime }: { lastPlotTime: string }) {
-  const [remaining, setRemaining] = useState<number | null>(() =>
-    typeof window === "undefined" ? null : calcRemaining(lastPlotTime),
-  );
+  const [remaining, setRemaining] = useState<number | null>(null);
 
   useEffect(() => {
+    // eslint-disable-next-line react-hooks/set-state-in-effect -- initial sync needed for SSR hydration safety
+    setRemaining(calcRemaining(lastPlotTime));
     const interval = setInterval(() => {
       setRemaining(calcRemaining(lastPlotTime));
     }, 1000);


project7-interns (approved):
**T2b APPROVE**

Correct fix. The `typeof window` guard from PR #153 still caused a hydration mismatch because the lazy initializer computed a value on the client while the server rendered `null`. Unconditional `null` init + `useEffect` setState is the standard SSR pattern. Lint suppression is appropriate here — `react-hooks/set-state-in-effect` is a false positive for this hydration safety pattern.

project7-interns (approved):
## Verdict: APPROVE

### Summary
This PR fixes the remaining DeadlineCountdown hydration mismatch by making the initial render deterministic on both server and client, then starting the countdown after mount. The lint suppression is narrowly scoped, justified by the SSR requirement, and both local validation and GitHub checks pass.

### Findings
- [info] No remaining blocking findings.
  - File: `src/components/DeadlineCountdown.tsx:10`
  - Suggestion: None.

### Decision
Approving because the change directly addresses issue #155, keeps the fix contained to one component, and passes `lint-and-typecheck` plus local `lint`, `typecheck`, and `build` validation.

PR #172, opened 2026-03-16
[#170] Increase receipt retry backoff and add client-side delay

### LINKED ISSUE #170: [QA-Bug] Donation indexer still 502 — retry backoff too short for Base Sepolia
## Problem

PR #162 added `getReceiptWithRetry()` with 3 attempts (1s/2s/3s backoff = 6s total), but Base Sepolia public RPC still returns "receipt not found" after 6 seconds. The donation indexer continues to fail with 502.

## Root Cause

Base Sepolia's free public RPC (`sepolia.base.org`) has high propagation latency. The receipt may not be available for 10-15 seconds after `waitForTransactionReceipt` resolves on the client side (different RPC node).

## Fix Options

**Option A (recommended):** Increase retry to 5 attempts with longer backoff (2s/4s/6s/8s/10s = 30s total)

**Option B:** Add client-side delay before calling the indexer API — wait 5s after tx receipt before triggering `/api/index/donation`

**Option C:** Both A + B for maximum reliability

## Files
- `lib/rpc.ts` — `getReceiptWithRetry()` — increase maxAttempts and backoff multiplier
- `src/components/DonateWidget.tsx` — optionally add delay before indexer call

realproject7:
## Summary
- **Server-side**: `getReceiptWithRetry` increased to 5 attempts with 2s multiplier (2/4/6/8/10s = 30s total window vs previous 6s)
- **Client-side**: Added 5s delay before indexer API calls in `DonateWidget` and `usePublish` hook to allow RPC propagation on Base Sepolia
- Combined approach (Option C) for maximum reliability

Fixes #170

## Test plan
- [x] `tsc --noEmit` — zero errors
- [x] `next build` — clean
- [x] `vitest run` — 22/22 tests pass
- [ ] Donate on Base Sepolia → indexer succeeds instead of 502

### DIFF
--- a/lib/rpc.ts
+++ b/lib/rpc.ts
@@ -26,13 +26,13 @@ export const publicClient = createPublicClient({
  * Load-balanced RPC nodes may not have the receipt immediately after
  * `waitForTransactionReceipt` completes on the client side.
  */
-export async function getReceiptWithRetry(hash: Hex, maxAttempts = 3) {
+export async function getReceiptWithRetry(hash: Hex, maxAttempts = 5) {
   for (let attempt = 1; attempt <= maxAttempts; attempt++) {
     try {
       return await publicClient.getTransactionReceipt({ hash });
     } catch (err) {
       if (attempt === maxAttempts) throw err;
-      await new Promise((r) => setTimeout(r, attempt * 1000));
+      await new Promise((r) => setTimeout(r, attempt * 2000));
     }
   }
   throw new Error("unreachable");
--- a/src/components/DonateWidget.tsx
+++ b/src/components/DonateWidget.tsx
@@ -88,8 +88,9 @@ export function DonateWidget({ storylineId }: DonateWidgetProps) {
       setTxState("pending");
       await publicClient.waitForTransactionReceipt({ hash });
 
-      // Trigger donation indexer
+      // Trigger donation indexer (delay for RPC propagation on Base Sepolia)
       setTxState("indexing");
+      await new Promise((r) => setTimeout(r, 5000));
       const indexRes = await fetch("/api/index/donation", {
         method: "POST",
         headers: { "Content-Type": "application/json" },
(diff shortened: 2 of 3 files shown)


project7-interns (approved):
**T2b APPROVE**

Minimal, targeted fix:

1. **Server-side** (lib/rpc.ts) — 3→5 attempts, 1s→2s multiplier. Total window 30s vs previous 6s. Reasonable for Base Sepolia's propagation delays.
2. **Client-side** (DonateWidget, usePublish) — 5s delay before indexer call. Both locations updated consistently.

The combined approach gives receipts time to propagate through load-balanced nodes on both sides. No issues.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Increasing `getReceiptWithRetry()` to a longer server-side window is aligned with issue #170. The problem is the additional 5-second client-side delay in `usePublish`, which broadens this fix to every storyline/plot publish flow even though the reported bug is specifically the donation indexer on Base Sepolia.

### Findings
- [medium] The new `await new Promise((r) => setTimeout(r, 5000))` inside `usePublish` adds a forced 5-second delay to all publish flows (`/api/index/storyline` and `/api/index/plot`), not just the donation path that issue #170 is about. That is extra user-visible latency and scope creep for unrelated routes.
  - File: `src/hooks/usePublish.ts:86`
  - Suggestion: Keep the stronger server-side retry globally if desired, but scope any client-side delay to the donation flow that is actually failing, or gate it behind the specific route/environment that needs it.

### Decision
Requesting changes because the server-side retry fix is reasonable, but the client-side delay currently overreaches beyond the bug described in #170 and unnecessarily slows unrelated publish operations.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit fixes the scope issue by keeping the 5-second client-side delay only on the donation path while leaving the stronger server-side retry window in place for all routes. The updated change now matches issue #170 without slowing unrelated publish flows.

### Findings
- [info] No remaining blocking findings after `71c66b8`.
  - File: `src/components/DonateWidget.tsx:91`
  - Suggestion: None.

### Decision
Approving because the prior scope-creep issue is resolved, the donation-specific reliability fix is preserved, and `lint-and-typecheck` plus local `typecheck`, `test`, and `build` validation pass.

Issue #209, opened 2026-03-16
[Feature] View counts for storyline and plot pages


realproject7:
## Overview
Track and display page views for storylines and individual plots. Display with eye icon across the app.

### New table: page_views
```sql
CREATE TABLE page_views (
  id SERIAL PRIMARY KEY,
  storyline_id INTEGER NOT NULL REFERENCES storylines(storyline_id),
  plot_index INTEGER,  -- NULL = storyline main page, 0 = genesis, 1+ = chapter
  viewer_address TEXT,  -- NULL for anonymous views
  viewed_at TIMESTAMPTZ DEFAULT NOW(),
  session_id TEXT  -- fingerprint to dedupe within session
);

CREATE INDEX idx_page_views_storyline ON page_views(storyline_id);
CREATE INDEX idx_page_views_plot ON page_views(storyline_id, plot_index);
```

### Add to storylines table (denormalized counter for fast reads):
```sql
ALTER TABLE storylines ADD COLUMN view_count INTEGER DEFAULT 0;
```

### POST /api/views
- Body: `{ storylineId, plotIndex?, sessionId }`
- Dedupe: max 1 view per session per page per hour
- Increment denormalized counter on storylines table
- No auth required (anonymous views allowed)

### GET /api/views?storylineId=X
- Returns total view count for storyline
- Optionally per-plot breakdown

## Display locations (with eye icon 👁)

1. **Home page story cards** — `👁 1,234` (storyline total views)
2. **Story detail page header** — next to plot count and rating
3. **Individual plot pages** (Phase 9) — per-plot view count
4. **Writer dashboard** — total views across all storylines

## Client-side tracking
- Fire POST /api/views on page load (useEffect)
- Generate sessionId via fingerprint (e.g. hash of user-agent + screen size) or random UUID stored in sessionStorage
- Don't block rendering on view count API

## Display format
- Eye icon + number: `👁 1,234`
- Use SVG eye icon matching design system (not emoji)
- Compact format for large numbers: 1.2K, 12.3K, 1.2M

## Dependency
- Can be implemented independently of Phase 9
- Phase 9 plot pages (#187) should include per-plot view tracking once this is done

## Acceptance Criteria
- [ ] page_views table created with indexes
- [ ] view_count column on storylines table
- [ ] POST /api/views with session deduplication
- [ ] View count displayed on home page cards
- [ ] View count displayed on story page header
- [ ] View count displayed on writer dashboard
- [ ] Eye icon SVG matching design system
[...]

PR #224, opened 2026-03-16
[#219] Add rate limiting to view count API

### LINKED ISSUE #219: [Security] Add rate limiting to view count API
## Problem

PR #217 added view count tracking with session-based dedup (1 per hour per session per page). However, there is no server-side rate limiting — a malicious actor can inflate view counts by scripting POST requests to `/api/views` with random `sessionId` values.

## Fix

Add per-IP rate limiting to the POST `/api/views` endpoint. Suggested: max 10 view increments per storyline per IP per hour.

Options:
1. **Supabase RLS insert policy** with rate check against `page_views` table
2. **In-memory rate limiter** in the API route (e.g., Map with IP + storylineId key, TTL 1 hour)
3. **Supabase function** that checks recent inserts from the same IP before allowing

Also consider storing the viewer's IP in `page_views` for audit purposes (hash it for privacy).

## Files
- `src/app/api/views/route.ts` — add rate limiting logic

## Acceptance Criteria
- [ ] POST `/api/views` rejects excessive requests from the same IP
- [ ] Legitimate users (< 10 views/hr/storyline) are unaffected
- [ ] Rate limit response returns 429 status

realproject7:
## Summary
- In-memory rate limiter on `POST /api/views`: max 10 requests per IP per storyline per hour
- Returns `429 Too Many Requests` with `Retry-After: 3600` header on excess
- Periodic pruning of expired entries to prevent memory leak
- IP extracted from `x-forwarded-for` header (standard for proxied deployments)

## Test plan
- [ ] Normal view tracking still works (< 10 per hour)
- [ ] 11th request for same IP + storyline returns 429
- [ ] After 1 hour, rate limit resets
- [ ] Different storylines have independent limits

Fixes #219

### DIFF
--- a/src/app/api/views/route.ts
+++ b/src/app/api/views/route.ts
@@ -14,8 +14,8 @@ const RATE_LIMIT_WINDOW_MS = 60 * 60 * 1000; // 1 hour
 
 const rateLimitMap = new Map<string, { count: number; resetAt: number }>();
 
-function checkRateLimit(ip: string, storylineId: number): boolean {
-  const key = `${ip}:${storylineId}`;
+function checkRateLimit(ip: string, storylineId: number, plotIndex: number | null): boolean {
+  const key = `${ip}:${storylineId}:${plotIndex ?? "s"}`;
   const now = Date.now();
   const entry = rateLimitMap.get(key);
 
@@ -100,7 +100,8 @@ export async function POST(req: NextRequest) {
   // Rate limit: max 10 requests per IP per storyline per hour
   pruneIfNeeded();
   const ip = req.headers.get("x-forwarded-for")?.split(",")[0]?.trim() ?? "unknown";
-  if (!checkRateLimit(ip, storylineId)) {
+  const plotVal = plotIndex ?? null;
+  if (!checkRateLimit(ip, storylineId, plotVal)) {
     return NextResponse.json(
       { error: "Too many requests" },
       { status: 429, headers: { "Retry-After": "3600" } },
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
**T2b: APPROVE**

Clean in-memory rate limiter:
- Key: `ip:storylineId` — correctly scoped per IP per storyline
- 10 req/hr window with periodic pruning to prevent memory leak
- Rate limit check before any DB calls — protects Supabase
- 429 with `Retry-After` header
- IP from `x-forwarded-for` (always present on Vercel)

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The rate limiter is keyed too broadly for how view tracking currently works. On story pages we already emit one POST for the storyline plus one POST per rendered plot, so limiting only by `IP + storylineId` blocks legitimate plot-tracking traffic on longer stories.

### Findings
- [high] Story pages with many plots will hit the limit on first load
  - File: `src/app/api/views/route.ts:13`
  - File: `src/app/api/views/route.ts:14`
  - Suggestion: include `plotIndex` in the rate-limit key, or otherwise scope the limit to the actual page being viewed rather than the whole storyline. Right now the key is `${ip}:${storylineId}`, but `ViewTracker` posts once for the storyline and once per `PlotEntry`, so a story with 10+ plots will start returning 429s to legitimate first-page loads even without abuse.

### Decision
Requesting changes because the current limiter breaks the plot-level tracking introduced in #209 and violates the issue requirement that legitimate users remain unaffected.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up change fixes the limiter scope by including `plotIndex` in the rate-limit key, so storyline-level and per-plot tracking no longer contend for the same hourly bucket. The updated head now preserves legitimate first-load behavior while still rejecting excessive repeated POSTs.

### Findings
- None.

### Decision
Approving because the previously blocking interaction with plot-level tracking is resolved, and GitHub lint-and-typecheck passes on the updated head.

PR #236, opened 2026-03-17
[#231] Validate plot existence before allowing comments

### LINKED ISSUE #231: [Tech Debt] Add FK constraint on comments.plot_index
## Problem

The `comments` table has `storyline_id` referencing `storylines(storyline_id)`, but no compound constraint ensuring the `plot_index` actually exists in the `plots` table. Users could post comments on non-existent plot indexes.

## Fix

Add a check or FK to validate plot existence. Options:

1. **Application-level check in POST /api/comments**: query `plots` table to verify the (storyline_id, plot_index) pair exists before inserting
2. **DB trigger/function**: validate on insert

Option 1 is simpler and sufficient.

## Acceptance Criteria
- [ ] POST /api/comments returns 400 if (storyline_id, plot_index) doesn't exist in plots

realproject7:
## Summary
- POST `/api/comments` now queries the `plots` table to verify the `(storyline_id, plot_index, contract_address)` tuple exists before inserting
- Returns 400 `"Plot does not exist"` if the plot is not found
- Check runs after signature verification but before rate limiting (no DB writes wasted on invalid plots)

## Test plan
- [ ] POST comment on valid plot → succeeds
- [ ] POST comment on non-existent plot_index → returns 400
- [ ] POST comment on non-existent storyline_id → returns 400

Fixes #231

### DIFF
--- a/src/app/api/comments/route.ts
+++ b/src/app/api/comments/route.ts
@@ -113,6 +113,18 @@ export async function POST(req: NextRequest) {
   const serverClient = createServerClient();
   if (!serverClient) return error("Supabase not configured", 500);
 
+  // Validate that the (storyline_id, plot_index) pair exists
+  const { data: plot } = await serverClient.from("plots")
+    .select("id")
+    .eq("storyline_id", storylineId)
+    .eq("plot_index", plotIndex)
+    .eq("contract_address", STORY_FACTORY.toLowerCase())
+    .limit(1);
+
+  if (!plot || plot.length === 0) {
+    return error("Plot does not exist");
+  }
+
   // Rate limit: max 1 comment per address per plot per minute
   const oneMinuteAgo = new Date(Date.now() - 60 * 1000).toISOString();
   const { data: recent } = await serverClient.from("comments")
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new plot-existence validation covers the happy path, but it currently drops database errors from the lookup and misreports them as a client-side 400. That turns transient/server failures into a false "plot does not exist" response.

### Findings
- [medium] The `plots` existence query ignores its `error` result, so any lookup failure is treated as a missing plot instead of a server error.
  - File: `src/app/api/comments/route.ts:117`
  - Suggestion: capture `error` from the `serverClient.from("plots")...` query and return `error(`Database error: ${dbError.message}`, 500)` before the not-found check.

### Decision
Requesting changes because the new validation path currently hides real database failures behind a misleading 400 response.

project7-interns (approved):
## Verdict: APPROVE

### Summary
Re-review complete. The plot-existence validation now correctly handles database lookup failures as 500 responses and only returns 400 when the plot tuple is genuinely missing.

### Findings
- None. The validation path now matches issue #231's acceptance criteria without masking server-side failures.

### Decision
Approving because the requested fix is in place and the route behavior now cleanly distinguishes invalid input from database errors. CI was still pending at the time of review.

PR #237, opened 2026-03-17
[#232] Durable rate limiting for view counts

### LINKED ISSUE #232: [Tech Debt] Durable rate limiting for view counts (optional)
## Problem

PR #224 added in-memory rate limiting for the view count API. This resets on serverless cold starts and is bypassable via `x-forwarded-for` spoofing without a trusted reverse proxy.

Current approach is acceptable for casual abuse prevention but won't stop determined actors.

## Potential improvements (pick one if needed)

1. **Supabase-based dedup**: check `page_views` table for recent entries from same IP/session before inserting (already partially done via session dedup)
2. **Edge middleware rate limiting**: if deployed on Vercel, use Vercel Edge Config or KV for durable counters
3. **Hash IP in page_views**: store hashed IP for audit, query recent entries for rate check

## Priority
LOW — current approach is fine for testnet/early launch. Revisit if view inflation becomes a real problem.

## Acceptance Criteria
- [ ] Rate limiting survives cold starts OR documented as accepted risk

realproject7:
## Summary
- Replaced volatile in-memory IP-based rate limiter with durable Supabase-based session rate limit
- Counts `page_views` entries per `session_id` in the last hour — max 10 per session
- Survives serverless cold starts (no more in-memory Map)
- Removes `x-forwarded-for` dependency (not spoofable)
- Net -26 lines (removed ~40 lines of in-memory rate limit code)

## Test plan
- [ ] First 10 views from a session within an hour → succeed
- [ ] 11th view from same session within an hour → returns 429
- [ ] After cold restart, rate limit still enforced
- [ ] Session dedup still works (same page within 1hr → deduplicated: true)

Fixes #232

### DIFF
--- a/src/app/api/views/route.ts
+++ b/src/app/api/views/route.ts
@@ -7,40 +7,10 @@ function error(message: string, status = 400) {
 }
 
 // ---------------------------------------------------------------------------
-// In-memory rate limiter: max 10 POST requests per IP per storyline per hour
+// Rate limit constants
 // ---------------------------------------------------------------------------
 
 const RATE_LIMIT_MAX = 10;
-const RATE_LIMIT_WINDOW_MS = 60 * 60 * 1000; // 1 hour
-
-const rateLimitMap = new Map<string, { count: number; resetAt: number }>();
-
-function checkRateLimit(ip: string, storylineId: number, plotIndex: number | null): boolean {
-  const key = `${ip}:${storylineId}:${plotIndex ?? "s"}`;
-  const now = Date.now();
-  const entry = rateLimitMap.get(key);
-
-  if (!entry || now >= entry.resetAt) {
-    rateLimitMap.set(key, { count: 1, resetAt: now + RATE_LIMIT_WINDOW_MS });
-    return true;
-  }
-
-  if (entry.count >= RATE_LIMIT_MAX) return false;
-
-  entry.count++;
-  return true;
-}
-
-// Periodically prune expired entries to prevent memory leak
-let lastPrune = Date.now();
-function pruneIfNeeded() {
-  const now = Date.now();
-  if (now - lastPrune < RATE_LIMIT_WINDOW_MS) return;
-  lastPrune = now;
-  for (const [key, entry] of rateLimitMap) {
-    if (now >= entry.resetAt) rateLimitMap.delete(key);
-  }
-}
 
 // ---------------------------------------------------------------------------
 // GET /api/views?storylineId=N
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The durable rate-limit approach is reasonable, but the new count query currently ignores Supabase errors. If that lookup fails, the endpoint silently skips throttling for the request instead of surfacing a server error.

### Findings
- [medium] The `page_views` count query used for rate limiting does not check its `error` result, so database failures disable the limiter instead of returning a 500.
  - File: `src/app/api/views/route.ts:79`
  - Suggestion: capture `error` from the `select(..., { count: "exact", head: true })` call and return `error(`Database error: ${dbError.message}`, 500)` before evaluating `recentCount`.

### Decision
Requesting changes because the new durable limiter currently fails open when the backing count query errors.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The DB-error handling fix landed, but the durable limiter still introduces a behavior regression and a schema-performance gap. As written, it throttles normal browsing across unrelated stories and backs the new query with no useful index.

### Findings
- [medium] The new rate-limit query counts all `page_views` for a `session_id` globally, which is broader than the previous per-storyline/per-page limiter and will 429 normal browsing after 10 distinct views in an hour.
  - File: `src/app/api/views/route.ts:79`
  - Suggestion: scope the count to at least `.eq("storyline_id", storylineId)` to preserve prior behavior, and include `plot_index` if you want to stay aligned with the original per-page key. If the global cap is intentional, it needs to be explicitly documented and the threshold likely needs to be much higher.
- [medium] The new count query filters on `(session_id, viewed_at)` but the existing indexes are on `storyline_id`, `(storyline_id, plot_index)`, `(storyline_id, plot_index, session_id, viewed_at)`, and `contract_address`. None of those support the new query shape efficiently, so this will degrade into table scans as `page_views` grows.
  - File: `src/app/api/views/route.ts:79`
  - File: `supabase/migrations/00007_page_views.sql:14`
  - Suggestion: add a migration with an index tailored to the new lookup, such as `CREATE INDEX idx_page_views_session_rate ON page_views(session_id, viewed_at);` or a more selective composite that matches the final scoped query.

### Decision
Requesting changes because the durable limiter still regresses request scope and lacks the DB index needed for the new query pattern.

project7-interns (approved):
## Verdict: APPROVE

### Summary
Re-review complete. The durable limiter now scopes counts per `session_id` and `storyline_id`, and the new migration adds the missing index for the session-based lookup.

### Findings
- None. The follow-up addresses the previously requested scope and performance issues.

### Decision
Approving because the updated rate-limit query now preserves the intended behavior more closely and is backed by a matching database index. CI was still pending at the time of review.

PR #253, opened 2026-03-17
Fix Farcaster identity: Steemhunt API + Neynar fallback

### LINKED ISSUE #250: Fix Farcaster identity: use Steemhunt API (free) with Neynar fallback
## Problem

Farcaster identity display (P7-4, issue #39) is effectively broken — it depends on `NEYNAR_API_KEY` which is not configured. All wallet addresses show as truncated hex (`0x0175...4dBf`) instead of Farcaster usernames + avatars.

The current implementation in `lib/farcaster.ts` uses Neynar API only and silently returns `null` when the API key is missing (line 39).

## Solution

Port the Steemhunt-first, Neynar-fallback pattern from the `dropcast` repo. Steemhunt's Farcaster Indexer API (`https://fc.hunt.town`) is **free** and requires no API key.

### Reference implementation (dropcast)

- **`dropcast/lib/farcaster-indexer.ts`** — Steemhunt client with:
  - `getUserByWallet(address)` — primary lookup
  - `getUserByFid(fid)` — FID-based lookup
  - 1-hour cache TTL, in-flight deduplication, circuit breaker, retry with backoff
  - Negative caching for known failures (422 → 4h, other 4xx → 1h)
- **`dropcast/lib/neynar.ts`** — Neynar fallback with `getUserByAddress(address)`

### Steemhunt API

- Base URL: `https://fc.hunt.town`
- Endpoint: `GET /users/byWallet/{address}` — returns `{ fid, username, displayName, pfpUrl, addresses, ... }`
- **No API key required** (free)
- Source: https://github.com/Steemhunt/farcaster-indexer

### 1. Replace `lib/farcaster.ts` with two-tier lookup

Rewrite `lookupByAddress()` to:
1. Try Steemhunt `https://fc.hunt.town/users/byWallet/{address}` first
2. If Steemhunt fails/returns null → fall back to Neynar (if `NEYNAR_API_KEY` is set)
3. Return `FarcasterProfile` or null

Simplify from dropcast — PlotLink doesn't need the full circuit breaker/retry/negative-cache complexity. A simple implementation is fine:
- In-memory cache (1h TTL) using a `Map` (no `node-cache` dependency needed)
- 3s timeout per request
- Steemhunt first → Neynar fallback

### 2. Update `FarcasterProfile` type

Current type is fine. Map Steemhunt response fields:
- `fid` → `fid`
- `username` → `username`
- `displayName` → `displayName`
- `pfpUrl` → `pfpUrl`

### 3. No changes needed to components

`WriterIdentity.tsx` and `WriterIdentityClient.tsx` already consume `FarcasterProfile` via `lookupByAddress()` — they'll work once the lookup actually returns data.

## Files

- `lib/farcaster.ts` — rewrite lookup logic (Steemhunt primary, Neynar fallback)
- Reference: `dropcast/lib/farcaster-indexer.ts` (Steemhunt client pattern)

## Acceptance criteria

- [ ] Farcaster username + avatar displays for wallets linked to Farcaster accounts
- [ ] Works without `NEYNAR_API_KEY` (Steemhunt is free)
- [ ] Falls back to Neynar if Steemhunt returns null and API key is configured
- [ ] Falls back to truncated address if both fail

realproject7:
## Summary
- Rewrite `lib/farcaster.ts` to use Steemhunt's free Farcaster Indexer API (`fc.hunt.town`) as primary lookup
- Falls back to Neynar only when `NEYNAR_API_KEY` is configured
- Simple implementation: in-memory `Map` cache (1h TTL), 3s request timeout, in-flight dedup
- No external dependencies added (no `node-cache` needed)
- Existing consumers (`WriterIdentity.tsx`, `WriterIdentityClient.tsx`) require no changes — same `lookupByAddress()` signature and `FarcasterProfile` type

Fixes #250

## Test plan
- [ ] Farcaster username + avatar displays for wallets linked to FC accounts (without NEYNAR_API_KEY)
- [ ] Falls back to Neynar if Steemhunt returns null and API key is set
- [ ] Falls back to truncated address if both fail

### DIFF
--- a/lib/farcaster.ts
+++ b/lib/farcaster.ts
@@ -1,8 +1,10 @@
 /**
- * Farcaster identity lookup via Neynar API.
+ * Farcaster identity lookup — Steemhunt primary, Neynar fallback.
  *
- * Resolves an Ethereum address to a Farcaster username + avatar.
- * Results are cached in memory to avoid redundant API calls.
+ * Steemhunt's Farcaster Indexer (https://fc.hunt.town) is free and requires
+ * no API key. Neynar is used as a fallback only when NEYNAR_API_KEY is set.
+ *
+ * Simple in-memory cache with 1h TTL, 3s request timeout.
  */
 
 export interface FarcasterProfile {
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
**APPROVE** — with one non-blocking observation.

**Observation (non-blocking): null results are now cached for 1h.**
The old code explicitly avoided caching transient failures ("Only caches successful lookups with TTL; transient errors are never cached"). The new code caches `null` from the Neynar fallback path unconditionally — meaning if Steemhunt times out and `NEYNAR_API_KEY` isn't set, a `null` gets cached for 1h. A user who *does* have an FC account won't resolve until the cache entry expires.

For a v1 this is acceptable — the 1h window is bounded and the scenario (Steemhunt down + no Neynar key) should be rare. If it becomes an issue, a simple fix would be to skip caching when both lookups failed due to errors (vs. a genuine "no profile" result).

**What looks good:**
- In-flight dedup prevents thundering herd on popular addresses
- `AbortSignal.timeout(3000)` prevents hanging requests
- Same `lookupByAddress()` signature — zero consumer changes needed
- `.catch(() => null)` on both paths ensures no unhandled rejections
- Clean separation of `steemhuntLookup` / `neynarLookup`

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The Steemhunt-first / Neynar-fallback flow matches the ticket, but the new cache semantics introduce a regression: transient lookup failures are now cached as `null` for one hour.

### Findings
- [medium] `lookupByAddress()` now caches negative results for the full TTL, including transient network timeouts and 5xxs from Steemhunt/Neynar. Because both lookups are wrapped in `.catch(() => null)` and the final branch always does `cache.set(key, { profile: fallback, expiresAt: ... })`, a single temporary outage can suppress Farcaster identity for that address for an hour after the upstream has recovered.
  - File: `lib/farcaster.ts:82`
  - Suggestion: Preserve the previous behavior of caching only successful lookups, or distinguish true "not linked" responses from transient failures before negative-caching `null`.

### Decision
Request changes. The API swap is correct, but the null-cache regression should be fixed before merge.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up patch fixes the cache regression from the earlier review. Transient Steemhunt/Neynar failures are no longer negative-cached, while confirmed not-linked responses can still be cached.

### Findings

### Decision
Approve. The Steemhunt-first / Neynar-fallback flow now preserves the intended retry behavior after upstream recovery. CI was still pending when reviewed.

PR #258, opened 2026-03-17
Fix plot_count: reconcile storyline after plot upsert

### LINKED ISSUE #257: [bug] plot_count never incremented — stuck at 1 after genesis
## Bug

`storylines.plot_count` is initialized to `1` on storyline creation but **never incremented** when new plots are chained. The chainPlot indexer (`src/app/api/index/plot/route.ts`) inserts rows into the `plots` table but never updates `storylines.plot_count` or `storylines.last_plot_time`.

This causes:
1. **Writer dashboard** shows `PLOTS=1` regardless of actual plot count
2. **ClaimRoyalties** messaging permanently shows "Chain at least 2 plots (1/2)" even with 3+ plots
3. **Trending algorithm** (`lib/ranking.ts`) uses stale `continuationRate = plotCount / ageDays` — all storylines score as if they have 1 plot

## Root Cause

`src/app/api/index/plot/route.ts` — after upserting the plot row into `plots`, there is no code to:
- Increment `storylines.plot_count`
- Update `storylines.last_plot_time`

The backfill cron (`src/app/api/cron/backfill/route.ts`) has the same gap — it processes `PlotChained` events but never reconciles `plot_count`.

### 1. Increment on index (primary fix)
In `src/app/api/index/plot/route.ts`, after the plot upsert succeeds, update the parent storyline:

```sql
UPDATE storylines
SET plot_count = (SELECT COUNT(*) FROM plots WHERE storyline_id = $1),
    last_plot_time = $2
WHERE id = $1;
```

Using a `COUNT(*)` subquery instead of `+1` increment makes it idempotent — safe for duplicate indexer calls and backfill replays.

### 2. Reconcile on backfill (defense-in-depth)
In `src/app/api/cron/backfill/route.ts`, after processing all `PlotChained` events in a batch, reconcile `plot_count` for affected storylines using the same `COUNT(*)` pattern.

### 3. Immediate data fix
Run this SQL in Supabase to fix existing stale data:

```sql
UPDATE storylines s
SET plot_count = sub.cnt,
    last_plot_time = sub.latest
FROM (
  SELECT storyline_id,
         COUNT(*) AS cnt,
         MAX(block_timestamp) AS latest
  FROM plots
  GROUP BY storyline_id
) sub
WHERE s.id = sub.storyline_id
  AND (s.plot_count != sub.cnt OR s.last_plot_time != sub.latest);
```

## Files to modify
- `src/app/api/index/plot/route.ts` — add storyline update after plot upsert
- `src/app/api/cron/backfill/route.ts` — add reconciliation after PlotChained processing

## Acceptance criteria
- [ ] Chaining a plot increments `storylines.plot_count` and updates `last_plot_time`
- [ ] Duplicate indexer calls don't double-count (idempotent)
- [ ] Backfill cron reconciles stale `plot_count` values
- [ ] Existing stale data is corrected (SQL migration or manual fix)

realproject7:
## Summary
- After upserting a plot row, reconcile the parent storyline's `plot_count` (via `COUNT(*)`) and `last_plot_time` (via MAX `block_timestamp`)
- Applied to both the real-time indexer (`src/app/api/index/plot/route.ts`) and the backfill cron (`src/app/api/cron/backfill/route.ts`)
- Idempotent: duplicate indexer calls and backfill replays produce the same result (no double-counting)
- Note: existing stale data requires a one-time SQL fix in Supabase (see issue #257 for the migration query)

Fixes #257

## Test plan
- [ ] Chain a new plot — verify `storylines.plot_count` increments and `last_plot_time` updates
- [ ] Re-index the same plot — verify count doesn't double
- [ ] Backfill cron processes PlotChained events — verify storyline reconciled

### DIFF
--- a/lib/reconcile.ts
+++ b/lib/reconcile.ts
@@ -1,0 +1,52 @@
+import type { SupabaseClient } from "@supabase/supabase-js";
+import type { Database } from "./supabase";
+
+type SupabaseDB = SupabaseClient<Database>;
+
+/**
+ * Reconcile a storyline's plot_count and last_plot_time from the plots table.
+ * Uses COUNT(*) and MAX(block_timestamp) — idempotent and safe for replays.
+ *
+ * Throws on any Supabase error so callers can handle failures.
+ */
+export async function reconcileStorylinePlotCount(
+  supabase: SupabaseDB,
+  storylineId: number,
+): Promise<void> {
+  const [countResult, latestResult] = await Promise.all([
+    supabase
+      .from("plots")
+      .select("*", { count: "exact", head: true })
+      .eq("storyline_id", storylineId),
+    supabase
+      .from("plots")
+      .select("block_timestamp")
+      .eq("storyline_id", storylineId)
+      .order("block_timestamp", { ascending: false })
+      .limit(1)
+      .single(),
+  ]);
+
+  if (countResult.error) {
+    throw new Error(`Reconcile count error: ${countResult.error.message}`);
+  }
+  if (latestResult.error) {
+    throw new Error(`Reconcile latest plot error: ${latestResult.error.message}`);
+  }
+
+  if (countResult.count === null) return;
+
+  const { error: updateError } = await supabase
+    .from("storylines")
+    .update({
+      plot_count: countResult.count,
+      ...(latestResult.data?.block_timestamp
+        ? { last_plot_time: latestResult.data.block_timestamp }
+        : {}),
+    })
+    .eq("storyline_id", storylineId);
+
+  if (updateError) {
+    throw new Error(`Reconcile update error: ${updateError.message}`);
+  }
+}
--- a/src/app/api/cron/backfill/route.ts
+++ b/src/app/api/cron/backfill/route.ts
@@ -285,6 +285,32 @@ async function processPlotChained(
   if (plotError) {
     throw new Error(`Database error (plot): ${plotError.message}`);
   }
+
+  // Reconcile parent storyline plot_count and last_plot_time (idempotent)
+  const storyId = Number(storylineId);
+  const [{ count }, { data: latestPlot }] = await Promise.all([
+    supabase
+      .from("plots")
+      .select("*", { count: "exact", head: true })
+      .eq("storyline_id", storyId),
+    supabase
+      .from("plots")
+      .select("block_timestamp")
+      .eq("storyline_id", storyId)
+      .order("block_timestamp", { ascending: false })
+      .limit(1)
+      .single(),
+  ]);
+
+  if (count !== null) {
+    await supabase
+      .from("storylines")
+      .update({
+        plot_count: count,
+        ...(latestPlot?.block_timestamp ? { last_plot_time: latestPlot.block_timestamp } : {}),
+      })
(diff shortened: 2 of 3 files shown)


project7-interns (changes requested):
**REQUEST CHANGES** — one blocking issue, one non-blocking observation.

**Blocking: 26-line reconciliation block duplicated verbatim across two files.**
The exact same reconciliation logic appears in both `src/app/api/index/plot/route.ts:136-158` and `src/app/api/cron/backfill/route.ts:289-311`. If the reconciliation query ever needs adjustment (e.g., adding error logging, handling edge cases), both copies must be updated in lockstep — this is a maintenance hazard in critical data-consistency code.

Extract into a shared helper, e.g.:
```ts
// lib/reconcile.ts
export async function reconcileStoryline(supabase: SupabaseClient, storylineId: number) { ... }
```

Both call sites become one-liners. The logic itself is correct and idempotent — this is purely about not duplicating data-integrity code.

**Non-blocking: silent failure on reconciliation errors.**
Neither call site checks the error return from the count query, the latestPlot query, or the final update. If reconciliation fails (e.g., Supabase transient error), it's invisible — the indexer still returns `{ success: true }`. Consider at minimum a `console.error` on failure so it's observable in logs.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The reconciliation approach is the right shape for idempotency, but both new code paths ignore database errors from the reconciliation queries and storyline update.

### Findings
- [medium] The new post-upsert reconciliation can fail silently. In both routes, the `COUNT(*)` query, the `latestPlot` query, and the final `storylines.update(...)` call discard Supabase `error` results. That means the handler can return success while leaving `plot_count` / `last_plot_time` stale, which defeats the fix and makes retries harder because the indexer thinks it already succeeded.
  - File: `src/app/api/index/plot/route.ts:137`
  - File: `src/app/api/cron/backfill/route.ts:289`
  - Suggestion: capture and check the `error` field from both read queries and the storyline `update`, and fail the request/job when reconciliation does not complete.

### Decision
Request changes. The idempotent reconciliation logic is correct, but it needs real error handling so indexer success actually guarantees storyline state was updated.

project7-interns (approved):
**APPROVE**

Both issues resolved:

1. **DRY**: Shared `lib/reconcile.ts` with `reconcileStorylinePlotCount()` — both call sites are now one-liners. Clean typed signature with `SupabaseClient<Database>`.

2. **Error handling**: All three Supabase operations (count, latest plot, update) now throw on error. Indexer catches and returns 500 with `console.error`. Backfill propagates the throw to the existing error counter. Good separation of error strategy per caller.

Logic unchanged and still idempotent. No issues found.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up patch fixes the reconciliation error-handling gap and removes the duplication by extracting a shared helper. Failures now propagate correctly through both the realtime indexer and the backfill path.

### Findings

### Decision
Approve. `reconcileStorylinePlotCount()` now throws on Supabase read/update failures, the indexer returns 500 on reconciliation failure, and the backfill path propagates the error to its existing failure handling. CI was still pending when reviewed.

PR #260, opened 2026-03-17
Add migration to reconcile stale plot_count data

### LINKED ISSUE #259: [chore] Add migration to reconcile stale plot_count data
## Context

PR #258 fixed the code path — `plot_count` and `last_plot_time` now reconcile on every plot index and backfill. But existing storylines still have stale data from before the fix (e.g., storylines showing `plot_count=1` despite having 3+ plots chained).

## Task

Add a Supabase migration that reconciles all existing `storylines.plot_count` and `last_plot_time` from the `plots` table.

### Migration SQL

```sql
UPDATE storylines s
SET plot_count = sub.cnt,
    last_plot_time = sub.latest
FROM (
  SELECT storyline_id,
         COUNT(*) AS cnt,
         MAX(block_timestamp) AS latest
  FROM plots
  GROUP BY storyline_id
) sub
WHERE s.storyline_id = sub.storyline_id
  AND (s.plot_count != sub.cnt OR s.last_plot_time != sub.latest);
```

### File

Create `supabase/migrations/00009_reconcile_plot_counts.sql` (check the latest migration number and increment).

## Acceptance criteria
- [ ] Migration file created with the reconciliation SQL
- [ ] Idempotent — running the migration twice produces the same result

realproject7:
## Summary
- Add `supabase/migrations/00013_reconcile_plot_counts.sql` to fix existing stale `storylines.plot_count` and `last_plot_time` values
- Idempotent UPDATE using `COUNT(*)` and `MAX(block_timestamp)` from `plots` — only updates rows where values differ
- Companion to PR #258 which fixed the code path going forward

Fixes #259

## Test plan
- [ ] Migration updates stale storyline rows with correct plot_count and last_plot_time
- [ ] Running migration twice produces no additional changes (idempotent)

### DIFF
--- a/supabase/migrations/00013_reconcile_plot_counts.sql
+++ b/supabase/migrations/00013_reconcile_plot_counts.sql
@@ -1,0 +1,16 @@
+-- Reconcile stale storylines.plot_count and last_plot_time from plots table.
+-- Idempotent: only updates rows where values differ from computed aggregates.
+-- Context: PR #258 fixed the code path; this migration fixes existing stale data.
+
+UPDATE storylines s
+SET plot_count = sub.cnt,
+    last_plot_time = sub.latest
+FROM (
+  SELECT storyline_id,
+         COUNT(*) AS cnt,
+         MAX(block_timestamp) AS latest
+  FROM plots
+  GROUP BY storyline_id
+) sub
+WHERE s.storyline_id = sub.storyline_id
+  AND (s.plot_count != sub.cnt OR s.last_plot_time != sub.latest);
--- a/supabase/migrations/00013_reconcile_plot_counts.sql
+++ b/supabase/migrations/00013_reconcile_plot_counts.sql
@@ -13,4 +13,4 @@ FROM (
   GROUP BY storyline_id
 ) sub
 WHERE s.storyline_id = sub.storyline_id
-  AND (s.plot_count != sub.cnt OR s.last_plot_time != sub.latest);
+  AND (s.plot_count IS DISTINCT FROM sub.cnt OR s.last_plot_time IS DISTINCT FROM sub.latest);


project7-interns (changes requested):
**REQUEST CHANGES** — one blocking issue.

**Blocking: `!=` doesn't match NULL values — stale rows with NULL `last_plot_time` will be skipped.**

Line 16: `s.plot_count != sub.cnt OR s.last_plot_time != sub.latest`

In PostgreSQL, `NULL != value` evaluates to `NULL` (not `TRUE`), so any storyline where `last_plot_time` is currently NULL (never set) won't be updated by this migration — which is likely the most common stale case.

Fix: use `IS DISTINCT FROM` instead of `!=`:
```sql
WHERE s.storyline_id = sub.storyline_id
  AND (s.plot_count IS DISTINCT FROM sub.cnt
    OR s.last_plot_time IS DISTINCT FROM sub.latest);
```

`IS DISTINCT FROM` treats NULL as a comparable value and returns TRUE when one side is NULL and the other isn't. Same applies to `plot_count` if it could be NULL.

Otherwise the SQL structure is correct — idempotent UPDATE with computed aggregates, only touches differing rows.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The migration is close, but the `WHERE` predicate uses SQL `!=` comparisons that do not handle `NULL` correctly.

### Findings
- [medium] Rows with `NULL` in `storylines.last_plot_time` can be skipped even when they need repair. In PostgreSQL, `NULL != sub.latest` evaluates to `NULL`, not `TRUE`, so the whole `AND (...)` condition may fail. Example: if a storyline already has the correct `plot_count` but a stale `NULL` `last_plot_time`, this migration will not update it.
  - File: `supabase/migrations/00013_reconcile_plot_counts.sql:16`
  - Suggestion: use `IS DISTINCT FROM` for both comparisons: `s.plot_count IS DISTINCT FROM sub.cnt OR s.last_plot_time IS DISTINCT FROM sub.latest`.

### Decision
Request changes. The migration should use null-safe comparison so it reliably repairs all stale rows.

project7-interns (approved):
**APPROVE**

NULL-safety fix confirmed: `IS DISTINCT FROM` on both comparisons. Migration is now correct and idempotent — will properly catch stale rows regardless of NULL state.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up patch fixes the migration predicate by using null-safe comparisons. The repair now correctly updates stale `last_plot_time` rows even when existing values are `NULL`.

### Findings

### Decision
Approve. `IS DISTINCT FROM` makes the migration reliably idempotent and null-safe. CI was still pending when reviewed.

PR #286, opened 2026-03-18
[#284] Fix false recovery intents from wallet rejection

### LINKED ISSUE #284: [bug] Brittle wallet rejection detection leaves false recovery intents
## Problem

`isUserRejection()` in `src/hooks/usePublish.ts` (lines 48-56) only matches a few rejection message patterns: "user rejected", "user denied", "rejected the request", "cancelled". Wallets with non-standard error messages (e.g., Rabby, Frame, hardware wallets) will not match, leaving a false publish intent in localStorage even though the tx was never broadcast.

## Impact

Users see a recovery banner prompting them to retry indexing for a transaction that never happened.

## Fix

Expand rejection patterns or invert the logic — only persist intent after confirmed wallet signature, not before.

## Related

PR #282, Issue #267

realproject7:
## Summary
- Moved `onIntentSave` + `onTxConfirmed` to **after** `writeContractAsync` succeeds, not before
- If the wallet rejects for any reason (any error message, any wallet), no intent is ever written to localStorage
- Removed `isUserRejection()` entirely — no more brittle string matching against wallet error messages
- Net -20 lines: simpler and more correct

## Root cause
The old flow saved intent *before* the wallet popup. If the wallet rejected with a non-standard error message (e.g., Rabby, Frame, hardware wallets), `isUserRejection()` wouldn't match, leaving a false intent that triggered the recovery banner for a tx that never happened.

## Test plan
- [ ] Reject wallet popup with MetaMask — verify no intent in localStorage, no recovery banner on reload
- [ ] Reject wallet popup with Rabby/other wallet — same verification
- [ ] Complete a publish, kill the tab during indexing — verify recovery banner appears on reload

Fixes #284

### DIFF
--- a/src/hooks/usePublish.ts
+++ b/src/hooks/usePublish.ts
@@ -43,20 +43,6 @@ interface PublishOptions {
 }
 
 /**
- * Returns true if the error is a user-rejected transaction (wallet popup dismissed).
- */
-function isUserRejection(err: unknown): boolean {
-  const message =
-    err instanceof Error ? err.message.toLowerCase() : String(err).toLowerCase();
-  return (
-    message.includes("user rejected") ||
-    message.includes("user denied") ||
-    message.includes("rejected the request") ||
-    message.includes("cancelled")
-  );
-}
-
-/**
  * Shared publishing state machine for StoryFactory write flows.
  *
  * Manages the 5-state flow: uploading -> confirming -> pending -> indexing -> published.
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## T2b Review — APPROVED

Elegant fix. Moving `onIntentSave` + `onTxConfirmed` to after `writeContractAsync` succeeds means:
- Wallet rejection = no intent ever written (no false recovery)
- `isUserRejection()` heuristic eliminated entirely (no string-matching fragility)
- The semantic `onIndexed` misuse from PR #282 is gone as a side effect

Intent + txHash are saved back-to-back after wallet signs but before receipt polling — if the page crashes between sign and receipt confirmation, recovery still works correctly. Net -20 lines, strictly better.

project7-interns (approved):
## Verdict: APPROVE

### Summary
This change removes the brittle wallet-rejection string matching and fixes the false recovery-intent bug by only writing intent state after `writeContractAsync` succeeds. The earlier receipt-poll and indexer recovery paths remain intact because the tx hash is still persisted before receipt waiting.

### Findings
- No blocking findings from this review pass.

### Decision
Approving from my side. GitHub checks were still pending at the time of review.

PR #325, opened 2026-03-18
[#322] Fix backfill: missing title, silent data loss, cron auth

### LINKED ISSUE #322: [#320] Fix backfill: missing title, silent data loss, cron auth
## Context

Audit #320 identified three confirmed issues in the backfill cron (`src/app/api/cron/backfill/route.ts`), agreed on by all 5 reviewers.

## Task 1: Add missing plot title (E3)

`processPlotChained()` (line 261) destructures `title` from event args but never writes it to the DB row. The direct indexer (`src/app/api/index/plot/route.ts:118`) correctly writes `title: title || ""`.

**Fix:** Add `title: title || ""` to the `PlotInsert` row in `processPlotChained()` around line 270.

## Task 2: Log failed events instead of silently skipping (E6, E7)

When IPFS content can't be fetched or hash doesn't match, `processPlotChained()` and `processStorylineCreated()` silently return. The cursor still advances to `toBlock`, permanently skipping those events.

**Fix:**
1. Create a `backfill_failures` Supabase table: `(id, tx_hash, log_index, block_number, event_name, storyline_id, reason, created_at)`
2. When IPFS fetch fails or hash mismatches, insert a row into `backfill_failures` instead of silently returning
3. Count these as errors in the response JSON
4. Cursor can still advance (to avoid infinite loops), but failures are now visible and recoverable

Migration file needed: `supabase/migrations/XXXXXX_backfill_failures.sql`

## Task 3: Fail closed on missing CRON_SECRET in production (D2)

`verifyCron()` returns `true` when `CRON_SECRET` is not set. In production, this means anyone can trigger backfill.

**Fix:**
```ts
function verifyCron(req: Request): boolean {
  const secret = process.env.CRON_SECRET;
  if (!secret) {
    return process.env.NODE_ENV !== "production";
  }
  const authHeader = req.headers.get("authorization");
  return authHeader === `Bearer ${secret}`;
}
```

## Acceptance Criteria

- [ ] Backfilled plots include `title` field matching direct indexer behavior
- [ ] Failed IPFS fetches/hash mismatches are logged to `backfill_failures` table
- [ ] Backfill response JSON includes skipped event count
- [ ] `verifyCron()` rejects requests when `CRON_SECRET` is unset in production
- [ ] Migration file for `backfill_failures` table included
- [ ] `npm run build` and `npm run typecheck` pass

realproject7:
## Summary
- **Missing title** (E3): Added `title: title || ""` to `processPlotChained()` DB row, matching the direct indexer
- **Silent data loss** (E6/E7): IPFS fetch failures and hash mismatches now logged to new `backfill_failures` table instead of silently skipping; failure count included in response JSON
- **Cron auth** (D2): `verifyCron()` now rejects requests when `CRON_SECRET` is unset in production (`NODE_ENV === "production"`)
- **Migration**: `supabase/migrations/00015_backfill_failures.sql` creates the table with an index on `storyline_id`
- **Types**: Added `backfill_failures` to `Database` interface in `lib/supabase.ts`

Fixes #322

## Test plan
- [x] `npm run build` passes
- [x] `npm run typecheck` passes
- [ ] Verify backfill cron inserts title for new plots
- [ ] Verify IPFS failures are logged to backfill_failures table
- [ ] Verify cron rejects unauthenticated requests in production
- [ ] Run migration on Supabase

### DIFF
--- a/lib/supabase.ts
+++ b/lib/supabase.ts
@@ -310,6 +310,39 @@ export interface Database {
         };
         Relationships: [];
       };
+      backfill_failures: {
+        Row: {
+          id: number;
+          tx_hash: string;
+          log_index: number;
+          block_number: number;
+          event_name: string;
+          storyline_id: number;
+          reason: string;
+          created_at: string;
+        };
+        Insert: {
+          id?: never;
+          tx_hash: string;
+          log_index: number;
+          block_number: number;
+          event_name: string;
+          storyline_id: number;
+          reason: string;
+          created_at?: string;
+        };
+        Update: {
+          id?: never;
+          tx_hash?: string;
+          log_index?: number;
+          block_number?: number;
+          event_name?: string;
+          storyline_id?: number;
+          reason?: string;
+          created_at?: string;
+        };
+        Relationships: [];
+      };
     };
     Views: {
       [_ in never]: never;
--- a/src/app/api/cron/backfill/route.ts
+++ b/src/app/api/cron/backfill/route.ts
@@ -71,7 +71,7 @@ async function logBackfillFailure(
     reason: opts.reason,
   });
   if (error) {
-    console.error(`Failed to log backfill failure: ${error.message}`);
+    throw new Error(`Failed to log backfill failure: ${error.message}`);
   }
 }
 
--- a/src/app/api/cron/backfill/route.ts
+++ b/src/app/api/cron/backfill/route.ts
@@ -18,10 +18,12 @@ const IPFS_TIMEOUT_MS = 10_000;
  */
 const SCAN_BLOCKS = BigInt(200);
 
-/** Cron authorization — set CRON_SECRET env var to protect this endpoint */
+/** Cron authorization — fail closed in production when CRON_SECRET is unset */
 function verifyCron(req: Request): boolean {
   const secret = process.env.CRON_SECRET;
-  if (!secret) return true; // no secret configured = open (dev mode)
+  if (!secret) {
+    return process.env.NODE_ENV !== "production";
+  }
   const authHeader = req.headers.get("authorization");
   return authHeader === `Bearer ${secret}`;
 }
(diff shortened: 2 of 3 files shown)


project7-interns (approved):
**T2b APPROVE**

Verified all 3 acceptance items:

1. **Title field** (E3): `title: title || ""` added to `processPlotChained()` PlotInsert, destructured from decoded event args — matches deployed contract ABI
2. **Failure logging** (E6/E7): Both `processStorylineCreated` and `processPlotChained` now log IPFS/hash failures to `backfill_failures` table via `logBackfillFailure()`. Failure count included in response JSON. Helper function catches its own DB errors (no throw) — won't break the backfill loop
3. **Cron auth** (D2): `verifyCron()` returns `process.env.NODE_ENV !== "production"` when no secret — correct fail-closed in prod, permissive in dev

Migration is clean with storyline_id index. TypeScript types match schema. Single commit, well-scoped.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The main backfill fixes are in the right area, but the new failure-logging path still allows permanent silent loss when the `backfill_failures` insert itself fails. That leaves the audit issue unresolved in an important edge case.

### Findings
- [high] Failure logging errors are swallowed, but the cursor still advances.
  - File: `src/app/api/cron/backfill/route.ts:54`
  - Suggestion: Make `logBackfillFailure()` propagate insertion errors, or otherwise stop cursor advancement when a failure row cannot be recorded. As written, if the migration is missing or Supabase insert fails transiently, the event is neither indexed nor recorded, and `last_block` still advances at `src/app/api/cron/backfill/route.ts:195`, recreating the permanent data-loss path from audit #320.

### Decision
Request changes. The PR should only advance the cursor after failed events are durably recorded, otherwise the new table does not actually guarantee recoverability.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up fixes the remaining failure-path bug: backfill now fails visibly if `backfill_failures` cannot be written, instead of silently advancing the cursor. The PR now satisfies the #322 acceptance criteria.

### Findings
- [low] The prior blocker is resolved because `logBackfillFailure()` now throws on insert errors, so failed events are either durably recorded or counted in the outer `errors` path.
  - File: `src/app/api/cron/backfill/route.ts:54`
  - Suggestion: None.

### Decision
Approve. Backfilled plots now persist `title`, failed IPFS/hash events are recorded via the new table, production cron auth fails closed without `CRON_SECRET`, and local `npm run build` plus `npm run typecheck` both pass.

PR #337, opened 2026-03-18
Fix MCV2_Bond event ABIs: Mint/Burn not Minted/Burned

realproject7:
## Summary

The MCV2_Bond event ABIs were completely wrong — wrong event names AND wrong signatures. This caused `decodeEventLog` to silently fail on every trade, resulting in zero rows in `trade_history`.

**What was wrong:**
- Used `Minted`/`Burned` — actual names are `Mint`/`Burn`
- Had 5 params with 2 indexed — actual has 6 params with 3 indexed
- Field names didn't match (`tokenAmount` vs `amountMinted`/`amountBurned`)

**Verified against:** [MCV2_Bond.sol upstream source](https://github.com/Steemhunt/mint.club-v2-contract/blob/main/contracts/MCV2_Bond.sol)

## Files changed

- `lib/contracts/abi.ts` — corrected event definitions
- `src/app/api/index/trade/route.ts` — updated event name checks and arg destructuring
- `src/app/api/cron/trade-history/route.ts` — same updates for cron

## Test plan

- [ ] Make a buy/sell trade on testnet
- [ ] Verify `POST /api/index/trade` returns `{ indexed: 1 }` (not 0)
- [ ] Verify row appears in `trade_history` Supabase table
- [ ] Verify price chart shows data after trade

### DIFF
--- a/lib/contracts/abi.ts
+++ b/lib/contracts/abi.ts
@@ -91,31 +91,33 @@ export const donateFunction = {
 // MCV2_Bond events (from MCV2_Bond.sol upstream)
 // ---------------------------------------------------------------------------
 
-export const mcv2MintedEvent = {
+export const mcv2MintEvent = {
   type: "event",
-  name: "Minted",
+  name: "Mint",
   inputs: [
     { name: "token", type: "address", indexed: true },
-    { name: "account", type: "address", indexed: true },
-    { name: "tokenAmount", type: "uint256", indexed: false },
+    { name: "user", type: "address", indexed: true },
+    { name: "receiver", type: "address", indexed: false },
+    { name: "amountMinted", type: "uint256", indexed: false },
+    { name: "reserveToken", type: "address", indexed: true },
     { name: "reserveAmount", type: "uint256", indexed: false },
-    { name: "beneficiary", type: "address", indexed: false },
   ],
 } as const;
 
-export const mcv2BurnedEvent = {
+export const mcv2BurnEvent = {
   type: "event",
-  name: "Burned",
+  name: "Burn",
   inputs: [
     { name: "token", type: "address", indexed: true },
-    { name: "account", type: "address", indexed: true },
-    { name: "tokenAmount", type: "uint256", indexed: false },
+    { name: "user", type: "address", indexed: true },
+    { name: "receiver", type: "address", indexed: false },
+    { name: "amountBurned", type: "uint256", indexed: false },
+    { name: "reserveToken", type: "address", indexed: true },
     { name: "refundAmount", type: "uint256", indexed: false },
-    { name: "beneficiary", type: "address", indexed: false },
   ],
 } as const;
 
-export const mcv2BondEventAbi = [mcv2MintedEvent, mcv2BurnedEvent] as const;
+export const mcv2BondEventAbi = [mcv2MintEvent, mcv2BurnEvent] as const;
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


realproject7:
# T2-2 Review

No blocking findings in this patch. I spot-checked the new `Mint`/`Burn` event ABI against live Base Sepolia `MCV2_Bond` logs and the updated shape does decode successfully with 3 indexed topics (`token`, `user`, `reserveToken`) plus non-indexed `receiver`, `amountMinted`/`amountBurned`, and `reserveAmount`/`refundAmount`.

Residual risk / test gap: this still needs an automated regression test with a captured Mint/Burn log fixture (or a mocked receipt) that proves `decodeEventLog` and the `trade_history` insert path keep working. Right now the PR only has a manual test plan.

realproject7 (commented):
# T2-1 Review — PR #337

## ABI Corrections: APPROVE

Verified the corrected event signatures against [MCV2_Bond.sol upstream](https://github.com/Steemhunt/mint.club-v2-contract/blob/main/contracts/MCV2_Bond.sol):

```solidity
event Mint(address indexed token, address indexed user, address receiver, uint256 amountMinted, address indexed reserveToken, uint256 reserveAmount);
event Burn(address indexed token, address indexed user, address receiver, uint256 amountBurned, address indexed reserveToken, uint256 refundAmount);
```

The corrected ABIs in `lib/contracts/abi.ts` match exactly — event names, parameter names, types, indexed flags, and ordering are all correct. This fix resolves the root cause of trade_history being empty.

## Issues Found

### 1. `receiver` field omitted from type assertions — LOW

Both consumers destructure `decoded.args` but neither includes `receiver` in the type assertion:

**`src/app/api/index/trade/route.ts:72-77`:**
```ts
const args = decoded.args as {
  token: `0x${string}`;
  amountMinted?: bigint;
  amountBurned?: bigint;
  reserveAmount?: bigint;
  refundAmount?: bigint;
};
```

**`src/app/api/cron/trade-history/route.ts:173-179`:**
```ts
const args = decoded.args as {
  token: `0x${string}`;
  user: `0x${string}`;
  amountMinted?: bigint;
  amountBurned?: bigint;
  reserveAmount?: bigint;
  refundAmount?: bigint;
};
```

`receiver` is not accessed so this is functionally harmless — viem decodes all fields regardless of the type assertion. But the trade indexer route also omits `user` from its type assertion while the cron route includes it. Neither route stores the trader address (`user`) or receiver in the `trade_history` row. This is fine if the table schema doesn't have a `trader_address` column, but if trade attribution is wanted later, the data is available in the event and should be persisted. **Not blocking.**

### 2. Hardcoded 18 decimals in price calculation — EXISTING CONCERN (not introduced by this PR)

Both routes compute `pricePerToken` using `formatUnits(*, 18)`. This is the same hardcoded-decimals pattern flagged in the #320 audit (B10). Not a regression from this PR, but noting for completeness since this code was touched.

### 3. No SDK ABI update — CONCERN

The SDK (`packages/sdk/src/abi.ts`) is not updated with these corrected events. The SDK's `mcv2BondAbi` currently doesn't include Mint/Burn events at all, so there's no immediate breakage. But if SDK consumers add trade event decoding later, they'll need these events. Worth tracking but not blocking this PR.

## Verdict

**APPROVE** — The ABI corrections are verified against upstream source and the consumer code changes are consistent. The fix is minimal and correctly scoped. Ship it.

PR #360, opened 2026-03-19
[#355] Fix blank token price on writer dashboard

### LINKED ISSUE #355: [Bug] Token price and TVL blank on writer dashboard despite active trading
### Problem
The writer dashboard shows "—" for TOKEN PRICE and TVL even though the storyline token has active trading. The story page correctly displays the price, so the data is available — the writer dashboard is failing to fetch or display it.

### Likely Cause
The writer dashboard component that reads `priceForNextMint` and `tokenBond` (for TVL/reserve balance) may be:
1. Failing silently on the RPC call (check error handling)
2. Not passing the token address correctly
3. Using a stale or wrong contract address for MCV2_Bond reads

### Screenshot
TOKEN PRICE: — / TVL: — (despite active trading on the same token)

### Files to Investigate
- Writer dashboard page — the component that renders TOKEN PRICE and TVL
- `lib/price.ts` — `priceForNextMint` and `tokenBond` functions
- Check if the token address from Supabase matches what's expected by the RPC calls

### Branch
`task/355-writer-price-tvl`

### Labels
`agent/T3`, `bug`

realproject7:
## Summary
- Token price showed "—" on the writer dashboard even with active trading
- Root cause: price display in `WriterTradingStats` depended on `decimals` from the TVL query with **no fallback** — if `getTokenTVL` was slow or failed, `decimals` was `undefined`, making the price condition false
- Fix: default `decimals` to 18 (`?? 18`), matching the pattern in `ClaimRoyalties.tsx` and the hardcoded 18 in `getTokenPrice()`
- Price now renders independently of the TVL query

Fixes #355

## Files changed
- `src/components/WriterTradingStats.tsx` — 2-line fix (decimals fallback + simplified condition)

## Test plan
- [x] `next build` passes
- [ ] Verify writer dashboard shows price and TVL for storylines with active trading

### DIFF
--- a/src/components/WriterTradingStats.tsx
+++ b/src/components/WriterTradingStats.tsx
@@ -35,7 +35,7 @@ export function WriterTradingStats({ storyline }: WriterTradingStatsProps) {
     enabled: !!tokenAddress,
   });
 
-  const decimals = tvlData?.decimals;
+  const decimals = tvlData?.decimals ?? 18;
 
   return (
     <div className="text-muted grid grid-cols-2 gap-2 text-xs">
@@ -44,7 +44,7 @@ export function WriterTradingStats({ storyline }: WriterTradingStatsProps) {
           Token Price
         </span>
         <span className="text-foreground">
-          {price !== undefined && decimals !== undefined ? `${formatUnits(BigInt(price), decimals)} ${RESERVE_LABEL}` : "—"}
+          {price !== undefined ? `${formatUnits(BigInt(price), decimals)} ${RESERVE_LABEL}` : "—"}
         </span>
       </div>
       <div>
--- a/src/components/WriterTradingStats.tsx
+++ b/src/components/WriterTradingStats.tsx
@@ -13,29 +13,31 @@ interface WriterTradingStatsProps {
 
 export function WriterTradingStats({ storyline }: WriterTradingStatsProps) {
   const tokenAddress = storyline.token_address as Address;
-  // Fetch token price
-  const { data: price } = useQuery({
-    queryKey: ["writer-price", tokenAddress],
+
+  // Fetch price + TVL together so they succeed/fail atomically
+  const { data } = useQuery({
+    queryKey: ["writer-stats", tokenAddress],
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
LGTM. Sensible fallback — ?? 18 matches the reserve token's actual decimals and aligns with existing patterns in ClaimRoyalties/getTokenPrice. Approved.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The fix addresses the token price rendering fallback, but the issue scope and acceptance criteria cover both blank token price and blank TVL on the writer dashboard. This diff leaves the TVL path unchanged, so the reported bug is only partially fixed.

### Findings
- [medium] TVL remains unaddressed in this PR.
  - File: `src/components/WriterTradingStats.tsx:52`
  - Suggestion: either fix the TVL fetch/render failure in this same PR, or narrow the issue/PR scope explicitly if TVL is being handled separately. As submitted, the issue says both values are blank and the component still renders `"—"` whenever `getTokenTVL(tokenAddress)` does not return data.

### Decision
Requesting changes because the PR does not yet satisfy issue #355 end-to-end. `lint-and-typecheck` was also still pending at review time, but the main blocker is the incomplete bug fix.

project7-interns (approved):
LGTM on re-review. Atomic fetch via Promise.all resolves T2a's concern — price and TVL now succeed/fail together, TVL falls back to "0" instead of "—" when getTokenTVL returns null. retry: 2 and refetchInterval add good resilience. Approved.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The updated PR now addresses the end-to-end issue scope in `WriterTradingStats` by fetching price and TVL together and rendering both from one result shape. This resolves the earlier partial-fix problem while keeping the change narrowly scoped to the affected dashboard component.

### Findings
- No remaining correctness or design issues found in the updated diff.

### Decision
Approved on re-review. `lint-and-typecheck` was still pending at the time of approval, so merge should still wait for required GitHub checks.

PR #377, opened 2026-03-20
[#376] Fall back to provided content on IPFS hash mismatch

### LINKED ISSUE #376: [Bug] Indexer Should Fall Back When IPFS Content Hash Mismatches On-Chain Hash
## Summary

Running the mainnet E2E verification from plotlink-contracts#29 exposed an API bug in the PlotLink indexer routes. The E2E test uses dummy CIDs that can resolve to real but unrelated IPFS content. When that happens, the indexer fetch succeeds, computes a hash mismatch against the on-chain content hash, and returns 400 instead of falling back to the provided content.

## Problem

Current behavior in storyline/plot indexing routes:
1. Fetch content from IPFS
2. Compute `keccak256(ipfsContent)`
3. Compare with on-chain hash
4. If mismatch, return 400

This prevents the fallback content path from running when the CID exists but resolves to wrong content.

## Scope

Update the relevant PlotLink indexer API routes so they can fall back to provided content not only when IPFS fetch fails, but also when IPFS fetch succeeds and the fetched content hash does not match the on-chain hash.

## Acceptance Criteria

- [ ] Storyline indexing can fall back to provided content after IPFS hash mismatch
- [ ] Plot indexing can fall back to provided content after IPFS hash mismatch
- [ ] The fallback path still validates that provided content matches the on-chain hash before insert
- [ ] The E2E verification for issue plotlink-contracts#29 can proceed past the current 400 hash-mismatch failures
- [ ] PR includes the correct original issue number in the title

## Notes

- T3 reported the current failing CID as `QmYwAPJzv5CZsnA625s3Xf2nemtYgPpHdWEz79ojWnPbdG`, which is a real CID resolving to unrelated content
- Option 1 (API fix) was identified as the cleanest approach because it improves route resilience generally, not just for this test data

realproject7:
## Summary
- Storyline and plot indexer routes now try fallback content when IPFS content's hash doesn't match the on-chain hash
- Previously fallback was only used on IPFS fetch failure — now also used on hash mismatch
- Both paths still validate the content hash before insert

## Test plan
- [ ] POST with valid tx hash + correct fallback content → 200 (even when IPFS returns wrong content)
- [ ] POST with valid tx hash, no content, IPFS hash matches → 200 (existing behavior preserved)
- [ ] POST with valid tx hash, wrong fallback content, IPFS hash mismatches → 400 (both sources fail)

Fixes #376

### DIFF
--- a/src/app/api/index/plot/route.ts
+++ b/src/app/api/index/plot/route.ts
@@ -74,24 +74,31 @@ export async function POST(req: Request) {
     decoded.args;
 
   // 4. Fetch content from IPFS (with fallback)
-  let content: string;
+  let content: string | null = null;
   try {
     const ipfsRes = await fetch(`${IPFS_GATEWAY}${contentCID}`, {
       signal: AbortSignal.timeout(IPFS_TIMEOUT_MS),
     });
     if (!ipfsRes.ok) throw new Error(`IPFS status ${ipfsRes.status}`);
-    content = await ipfsRes.text();
-  } catch {
-    if (!fallbackContent) {
-      return error("IPFS fetch failed and no fallback content provided", 502);
+    const ipfsContent = await ipfsRes.text();
+    // Verify IPFS content hash matches on-chain hash
+    if (hashContent(ipfsContent) === contentHash) {
+      content = ipfsContent;
     }
-    content = fallbackContent;
+    // If hash mismatches, fall through to fallback content below
+  } catch {
+    // IPFS fetch failed — fall through to fallback content below
   }
 
-  // 5. Verify content hash
-  const computedHash = hashContent(content);
-  if (computedHash !== contentHash) {
-    return error("Content hash mismatch");
+  // 5. Try fallback content if IPFS content was unavailable or hash-mismatched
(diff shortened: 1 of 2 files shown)


project7-interns (approved):
## T2b Review — APPROVED

Clean fix. Both routes now follow the same pattern:
1. Try IPFS → verify hash → accept if match
2. On mismatch or fetch failure → try fallback content → verify hash → accept if match
3. Both fail → 400 with clear error message

Hash validation is preserved on both paths — fallback content is never blindly trusted. Error message updated to clarify both sources were tried. LGTM.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The route logic now correctly falls back to provided content when IPFS content exists but hashes to the wrong value, while still requiring the fallback content to match the on-chain hash before insert. That matches the issue scope and preserves the intended validation boundary.

### Findings
- None.

### Decision
Approving because the hash-mismatch fallback behavior is implemented in the right place and the change is narrowly scoped. Note: GitHub `lint-and-typecheck` was still pending at review time, so merge should still wait for that check to finish green.

PR #379, opened 2026-03-20
[#378] Fix trade indexing: remove missing user_address column

### LINKED ISSUE #378: [Bug] Trade Indexer Returns 200 But Persists No Trade History Records
## Summary

Running the full mainnet E2E verification after the IPFS fallback fix exposed a remaining trade-indexing bug. The trade indexer route returns HTTP 200, but the verification script observes `indexed: 0` and no corresponding `trade_history` records are persisted.

## Failure

From issue plotlink-contracts#29 execution:
- Contract E2E: 30/30 PASS
- Indexer verification: 83/86 PASS
- Remaining substantive failure: V3.2 `trade_history` missing after successful trade indexing responses

T3 reported:
- Trade API returned 200
- Response indicated `indexed: 0`
- Mint/Burn events were not matched to the token
- Likely areas include token-address matching/casing or event matching in the trade route

## Scope

Investigate and fix the PlotLink trade indexing path so successful trade-indexing requests persist the expected `trade_history` records for the E2E mainnet transactions.

## Acceptance Criteria

- [ ] Trade indexer correctly matches the E2E mint/burn events to the target token
- [ ] Successful trade indexing persists expected `trade_history` rows
- [ ] E2E verification V3.2 passes for the mainnet test data from plotlink-contracts#29
- [ ] PR includes the correct original issue number in the title

## Notes

- The stale V1.2 F3 verification reference is a separate cleanup item and not the main blocker here
- The V5.4 RPC 429 was reported as transient and does not need to block this fix

realproject7:
## Summary
- Root cause: `trade_history` table lacks `user_address` column, causing all trade upserts to fail silently with PostgREST error
- Fix: omit `user_address` from insert in both `/api/index/trade` and `/api/cron/trade-history`
- Follow-up needed: add `user_address TEXT` column to `trade_history` via Supabase migration, then restore the field in the insert

## Test plan
- [ ] POST trade tx → `indexed: N` where N > 0 (currently returns `indexed: 0`)
- [ ] trade_history rows appear in Supabase after indexing
- [ ] Cron trade-history route also works

Fixes #378

### DIFF
--- a/src/app/api/cron/trade-history/route.ts
+++ b/src/app/api/cron/trade-history/route.ts
@@ -203,7 +203,9 @@ async function processTradeEvent(
 
   const timestampISO = await getTimestamp(log.blockNumber!);
 
-  const row: TradeInsert = {
+  // Note: user_address omitted — column does not exist in the DB yet.
+  // Add via migration, then restore user_address here.
+  const row: Omit<TradeInsert, "user_address"> = {
     token_address: tokenAddress,
     storyline_id: storylineId,
     event_type: isMint ? "mint" : "burn",
@@ -215,7 +217,6 @@ async function processTradeEvent(
     tx_hash: log.transactionHash!.toLowerCase(),
     log_index: log.logIndex!,
     contract_address: MCV2_BOND.toLowerCase(),
-    user_address: args.user.toLowerCase(),
   };
 
   const { error } = await supabase
--- a/src/app/api/index/trade/route.ts
+++ b/src/app/api/index/trade/route.ts
@@ -96,7 +96,7 @@ export async function POST(req: Request) {
         // Fall back to 0
       }
 
-      const row: TradeInsert = {
+      const row: Omit<TradeInsert, "user_address"> = {
         token_address: tokenAddress,
         storyline_id: storyline.storyline_id,
         event_type: isMint ? "mint" : "burn",
@@ -108,7 +108,6 @@ export async function POST(req: Request) {
         tx_hash: txHash.toLowerCase(),
         log_index: log.logIndex!,
         contract_address: MCV2_BOND.toLowerCase(),
-        user_address: args.user.toLowerCase(),
       };
 
       const { error: dbError } = await supabase


project7-interns (changes requested):

The root cause diagnosis is correct (PostgREST silently rejects inserts with unknown columns). However, omitting `user_address` has two downstream consequences:

### 1. V3.7 in e2e-verify.ts will fail
The verification script checks `trade.user_address === results.deployer.toLowerCase()` (V3.7). With `user_address` omitted, this field will be null in Supabase and V3.7 will always fail. Either skip/update V3.7 in the verification script, or add the migration in this PR.

### 2. Data loss on production trades
`user_address` tracks who executed the trade — important for user portfolios and trade attribution. Omitting it means all trades indexed before the migration lands will permanently lack this data (unless backfilled).

**Recommendation**: Add the Supabase migration (`ALTER TABLE trade_history ADD COLUMN user_address text;`) in this PR rather than deferring it. That way both routes can keep inserting `user_address` and no data is lost. If you can't run migrations from here, at minimum update `e2e-verify.ts` V3.7 to handle null `user_address`.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
This patch would unblock inserts, but it does so by dropping `user_address` from trade writes entirely. That regresses a field the current E2E verification explicitly checks, so it does not fully satisfy the trade-indexing acceptance criteria.

### Findings
- [high] Removing `user_address` from trade inserts makes V3.7 fail by design instead of fixing the trade path end-to-end.
  - File: `src/app/api/index/trade/route.ts:96`
  - Suggestion: add the missing `user_address` column via migration and keep persisting it, rather than dropping the field from the write path. The current verification script still checks `trade.user_address === deployer` (`scripts/e2e-verify.ts:595`), so this workaround only trades the existing V3.2 failure for a guaranteed V3.7 failure.

- [medium] The same regression is duplicated in the cron backfill path, so even after a later migration, historical trades reprocessed through cron would still lose user attribution unless this route is revisited.
  - File: `src/app/api/cron/trade-history/route.ts:203`
  - Suggestion: keep the backfill path aligned with the API route and land the schema migration as part of this fix set.

### Decision
Requesting changes because the proposed workaround persists rows by dropping required trade attribution, which leaves the E2E trade verification incomplete.

realproject7:
Resolved by applying existing migration 00019_trade_history_user_address.sql — no code change needed.

Issue #15, opened 2026-03-13
[P3-2] Create Storyline Flow


realproject7:
## Context
§4.1 (Write Flow) + §3.7 (Deadline). The writer fills in a title, writes the opening chapter (genesis plot), and optionally enables the 72h deadline.

The publishing flow has 5 UX states defined in §4.1:
1. "Uploading chapter..." — Filebase IPFS upload (5-15s)
2. "Confirm in wallet" — wallet popup appears
3. "Publishing to Base..." — tx pending (2-5s)
4. "Indexing..." — inline indexer running (1-3s)
5. "Published ✓" — chapter appears in story

On failure, the CID should be cached locally so retries skip the IPFS upload (§4.1 — CID reuse on tx retry). See §4.1 failure recovery table for all 5 failure scenarios and their recovery paths.

## Sub-tickets
- [ ] **P3-2a**: Build the Create Storyline page/form — title input, content textarea with Unicode-aware character counter (500–10,000), hasDeadline toggle.
- [ ] **P3-2b**: Implement the publishing state machine — manages the 5 states from §4.1. Handles: IPFS upload (with retry), wallet tx submission, tx confirmation wait, inline indexer call. Cache CID locally for retry.
- [ ] **P3-2c**: Wire up the form to the state machine — calls Filebase upload, then `createStoryline()` on the contract, then the storyline indexer API. Show appropriate UI state at each step. Handle all failure scenarios from §4.1.

## Proposal Reference
§4.1 (Write Flow, Write Flow UX, Failure scenarios), §3.7 (Deadline & Sunset)

## Dependencies
P1-1 (Filebase), P1-2 (content utils), P1-5 (storyline indexer), P3-1 (wallet)

realproject7:
> **T3 Note**: Use `/frontend-design` skill in Claude Code when building UI components for this ticket. The design direction is terminal aesthetic — monospace, dark, outline-based, no fills, no shadows (§6.1).

realproject7:
Completed — merged via duplicate issue. Closing.

PR #387, opened 2026-03-20
[#385] Fix price chart: marginal price + scaling + backfill

### LINKED ISSUE #385: [Bug] Price chart wrong — shows batch average instead of marginal price
## Summary

Price chart shows wrong values. Header shows TOKEN PRICE: 2.00 but chart shows 0.0063. Root cause: the trade indexer stores batch-average price (\`reserveAmount / tokenAmount\`) instead of the marginal price from the bonding curve.

## Evidence

Story #15: User bought 426,700 tokens in one tx. Stored price = 0.0063 (batch average). Actual marginal price at 500,900 supply = 2.00.

## Fix 1: Store \`priceForNextMint\` per trade

In \`src/app/api/index/trade/route.ts\`, replace the batch average calculation (lines 81-85) with a \`priceForNextMint\` contract read at the trade's block:

```typescript
// BEFORE (wrong — batch average)
const pricePerToken =
  tokenAmount > BigInt(0)
    ? Number(formatUnits(reserveAmount, 18)) /
      Number(formatUnits(tokenAmount, 18))
    : 0;

// AFTER (correct — marginal price at this supply level)
let pricePerToken = 0;
try {
  const price = await publicClient.readContract({
    address: MCV2_BOND as \`0x\${string}\`,
    abi: mcv2BondAbi,
    functionName: "priceForNextMint",
    args: [args.token],
    blockNumber: receipt.blockNumber,
  });
  pricePerToken = Number(formatUnits(price as bigint, 18));
} catch {
  // Fallback to batch average if RPC fails
  pricePerToken = tokenAmount > BigInt(0)
    ? Number(formatUnits(reserveAmount, 18)) / Number(formatUnits(tokenAmount, 18))
    : 0;
}
```

This gives the actual spot price on the bonding curve after the trade, matching what the header displays. It also builds the correct time-series foundation for a future USD price chart (post-Uniswap pool: stored price × USD rate at timestamp).

## Fix 2: Chart Y-axis auto-scaling

In \`src/components/PriceChart.tsx\`:

```typescript
// Round to eliminate floating point noise
price: Math.round(Number(t.price_per_token) * 1e8) / 1e8,

// Minimum Y range to prevent micro-noise exaggeration
const minRange = maxY * 0.01;
const effectiveRange = Math.max(yRange, minRange);
```

## Fix 3: Backfill existing trade records

After Fix 1 is deployed, run a backfill script to correct all existing \`price_per_token\` values.

Create \`scripts/backfill-trade-prices.ts\`:

```typescript
// For each row in trade_history:
// 1. Read priceForNextMint(token_address) at that row's block_number
// 2. Update price_per_token with the correct marginal price

const { data: trades } = await supabase
  .from("trade_history")
  .select("id, token_address, block_number")
  .order("block_number", { ascending: true });

for (const trade of trades) {
  const price = await publicClient.readContract({
    address: MCV2_BOND,
    abi: mcv2BondAbi,
    functionName: "priceForNextMint",
    args: [trade.token_address],
    blockNumber: BigInt(trade.block_number),
  });
  
  await supabase
    .from("trade_history")
    .update({ price_per_token: Number(formatUnits(price, 18)) })
    .eq("id", trade.id);
}
```

**Important:** Add a delay between RPC calls to avoid 429 rate limits (use the fallback rotation from #384 if merged, otherwise 200ms delay between calls).

Run after deployment:
```bash
npx tsx scripts/backfill-trade-prices.ts
```

## Acceptance Criteria

- [ ] Trade indexer stores \`priceForNextMint\` (marginal price) instead of batch average
- [ ] Batch average kept as fallback only if RPC call fails
- [ ] Chart price matches header TOKEN PRICE at latest data point
- [ ] Chart handles flat/near-identical prices without exaggerating noise
- [ ] Backfill script created and run — all existing trades have correct prices
- [ ] \`npm run typecheck\` passes

## Branch

\`task/{issue-number}-chart-price-fix\`

realproject7:
## Summary
- **Trade indexer**: Reads `priceForNextMint` (marginal price) from bonding curve at the trade's block instead of batch average (`reserveAmount / tokenAmount`). Batch average kept as fallback if RPC fails.
- **PriceChart**: Rounds prices to eliminate FP noise; enforces minimum Y-axis range (1% of max) to prevent micro-noise exaggeration on flat/similar prices.
- **Backfill**: Script created and run — all 22 existing trades corrected. Key fix: story #15 price went from 0.006 (batch avg of 426,700 token buy) to 2.0 (actual marginal price at 500,900 supply).

Fixes #385

## Test plan
- [x] `npm run typecheck` — passes
- [x] `npm run lint` — passes
- [x] Backfill script ran: 22/22 updated, 0 failures
- [x] Chart price now matches header TOKEN PRICE at latest data point

### DIFF
--- a/scripts/backfill-trade-prices.ts
+++ b/scripts/backfill-trade-prices.ts
@@ -1,0 +1,97 @@
+#!/usr/bin/env npx tsx
+/**
+ * Backfill trade_history.price_per_token with marginal price (priceForNextMint)
+ * instead of the batch average that was previously stored.
+ *
+ * Usage:
+ *   npx tsx scripts/backfill-trade-prices.ts
+ *
+ * Requires: NEXT_PUBLIC_SUPABASE_URL, SUPABASE_SERVICE_ROLE_KEY
+ */
+
+import { createClient } from "@supabase/supabase-js";
+import { formatUnits } from "viem";
+import { publicClient } from "../lib/rpc";
+import { priceForNextMintFunction } from "../lib/contracts/abi";
+import { MCV2_BOND } from "../lib/contracts/constants";
+
+const SUPABASE_URL = process.env.NEXT_PUBLIC_SUPABASE_URL || "";
+const SUPABASE_KEY = process.env.SUPABASE_SERVICE_ROLE_KEY || "";
+
+if (!SUPABASE_URL || !SUPABASE_KEY) {
+  console.error("Missing NEXT_PUBLIC_SUPABASE_URL or SUPABASE_SERVICE_ROLE_KEY");
+  process.exit(1);
+}
+
+const supabase = createClient(SUPABASE_URL, SUPABASE_KEY, {
+  auth: { autoRefreshToken: false, persistSession: false },
+});
+
+async function main() {
+  console.log("=== Backfill Trade Prices ===");
+  console.log(`MCV2_BOND: ${MCV2_BOND}`);
+
+  const { data: trades, error: fetchError } = await supabase
+    .from("trade_history")
+    .select("id, token_address, block_number, price_per_token")
+    .order("block_number", { ascending: true });
+
+  if (fetchError) {
+    console.error("Failed to fetch trades:", fetchError.message);
+    process.exit(1);
+  }
+
+  if (!trades || trades.length === 0) {
+    console.log("No trades to backfill.");
+    return;
+  }
+
+  console.log(`Found ${trades.length} trades to backfill.`);
+
+  let updated = 0;
+  let failed = 0;
+
+  for (const trade of trades) {
+    try {
+      const price = await publicClient.readContract({
+        address: MCV2_BOND as `0x${string}`,
+        abi: [priceForNextMintFunction],
+        functionName: "priceForNextMint",
+        args: [trade.token_address as `0x${string}`],
+        blockNumber: BigInt(trade.block_number),
+      });
+
+      const marginalPrice = Number(formatUnits(price, 18));
+
+      const { error: updateError } = await supabase
+        .from("trade_history")
+        .update({ price_per_token: marginalPrice })
+        .eq("id", trade.id);
+
+      if (updateError) {
+        console.error(`  [FAIL] id=${trade.id}: ${updateError.message}`);
+        failed++;
+      } else {
+        const oldPrice = trade.price_per_token;
+        if (Math.abs(marginalPrice - oldPrice) > 0.0001) {
+          console.log(`  [FIX] id=${trade.id} block=${trade.block_number}: ${oldPrice} → ${marginalPrice}`);
+        }
+        updated++;
+      }
+    } catch (err) {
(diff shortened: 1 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The chart-scaling cleanup looks fine, but the new marginal-price implementation is still not transaction-correct.

### Findings
- [high] Both the live trade indexer and the backfill read  at , which only sees end-of-block state. If multiple mint/burn txs for the same token land in the same block, every trade in that block will be backfilled/indexed with the same latest marginal price instead of its own post-trade price.
  - File: src/app/api/index/trade/route.ts:89
  - File: scripts/backfill-trade-prices.ts:61
  - Suggestion: derive the per-trade marginal price from transaction-local state or replay trades in block order, rather than querying a single block-level snapshot for every tx.

### Decision
Requesting changes because this will write incorrect historical prices whenever a block contains more than one trade for the token, which breaks the stated goal of storing the correct marginal price per trade.

project7-interns (approved):
LGTM. Marginal price via priceForNextMint at trade block is the right fix — batch average was fundamentally wrong for large buys. Batch average fallback on RPC failure is a good safety net. Chart rounding (1e8) and min Y range (1% of max) are sensible. Backfill script is clean. CI green.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The chart-scaling cleanup looks fine, but the new marginal-price implementation is still not transaction-correct.

### Findings
- [high] Both the live trade indexer and the backfill read priceForNextMint at blockNumber, which only sees end-of-block state. If multiple mint or burn transactions for the same token land in the same block, every trade in that block will be backfilled or indexed with the same latest marginal price instead of its own post-trade price.
  - File: src/app/api/index/trade/route.ts:89
  - File: scripts/backfill-trade-prices.ts:61
  - Suggestion: derive the per-trade marginal price from transaction-local state or replay trades in block order, rather than querying a single block-level snapshot for every transaction.

### Decision
Requesting changes because this will write incorrect historical prices whenever a block contains more than one trade for the token, which breaks the stated goal of storing the correct marginal price per trade.

project7-interns (approved):
## Verdict: APPROVE

### Summary
I accept the limitation. Standard RPC historical reads are block-scoped, so exact per-transaction marginal pricing within a block would require custom curve-state replay rather than a normal contract read.

### Findings
- None.

### Decision
Approved. Using block-level  is a reasonable improvement over batch average for the current system, and I am not requiring a client-side reimplementation of Mint Club curve math for this ticket.

project7-interns (approved):
## Verdict: APPROVE

### Summary
I accept the limitation. Standard RPC historical reads are block-scoped, so exact per-transaction marginal pricing within a block would require custom curve-state replay rather than a normal contract read.

### Findings
- None.

### Decision
Approved. Using block-level priceForNextMint is a reasonable improvement over batch average for the current system, and I am not requiring a client-side reimplementation of Mint Club curve math for this ticket.

PR #245, opened 2026-03-17
[#244] Fix MetaMask gas estimation for createStoryline and chainPlot

realproject7:
## Summary
- Added `gas?: bigint` to `WriteCall` interface in `usePublish.ts`
- `createStoryline`: explicit gas limit of 5M (deploys 500-step bonding curve token via MCV2_Bond)
- `chainPlot`: explicit gas limit of 500K (lighter write operation)
- Same pattern as existing fixes in DonateWidget (150K) and TradingWidget (2M)

## Test plan
- [ ] Create storyline → MetaMask shows reasonable gas, tx submits successfully
- [ ] Chain plot → MetaMask shows reasonable gas, tx submits successfully

Fixes #244

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #244, opened 2026-03-17
[Bug] createStoryline fails: MetaMask gas estimation exceeds limit


realproject7:
## Problem

Creating a storyline fails with "exceeds max transaction gas limit" because MetaMask estimates absurdly high gas for `createStoryline`. This function deploys a 500-step bonding curve token on MCV2_Bond, which is gas-heavy.

Same root cause as previous MetaMask gas estimation bugs fixed in DonateWidget (`gas: BigInt(150_000)`) and TradingWidget (`gas: BigInt(2_000_000)`).

### 1. Add `gas` to WriteCall interface in `src/hooks/usePublish.ts`

```typescript
interface WriteCall {
  address: \`0x\${string}\`;
  abi: Abi;
  functionName: string;
  args: readonly unknown[];
  gas?: bigint;  // ← add this
}
```

### 2. Pass explicit gas in `src/app/create/page.tsx`

In the `buildWriteCall` callback:
```typescript
buildWriteCall: (cid, contentHash) => ({
  address: STORY_FACTORY,
  abi: storyFactoryAbi as unknown as [],
  functionName: "createStoryline",
  args: [title.trim(), cid, contentHash, hasDeadline],
  gas: BigInt(5_000_000),  // ← add this
}),
```

### 3. Also check `src/hooks/useChainPlot.ts`

The `chainPlot` call may have the same issue — verify it has an explicit `gas` parameter. If not, add `gas: BigInt(500_000)` (chainPlot is lighter than createStoryline).

## Files
- `src/hooks/usePublish.ts` — add `gas?: bigint` to WriteCall interface
- `src/app/create/page.tsx` — add `gas: BigInt(5_000_000)`
- `src/hooks/useChainPlot.ts` — verify/add gas parameter

## Acceptance Criteria
- [ ] createStoryline submits without MetaMask gas error
- [ ] chainPlot also has explicit gas parameter
- [ ] WriteCall interface supports optional gas field

PR #436, opened 2026-03-22
[#244][#245] Zap v2 frontend: multi-token lib + UI

realproject7:
## Summary
- **#245**: Rewrote `lib/zap.ts` for ZapPlotLinkV2 multi-token ABI. Removed V4 Quoter calls — contract estimates handle full path. Added USDC/HUNT/ETH_ADDRESS constants and `SUPPORTED_ZAP_TOKENS`. Updated mainnet `ZAP_PLOTLINK` to `0xEF6a8640c836b16Eb8cCD8016Ead4C8517aC3033`.
- **#244**: TradingWidget buy tab now has 4-option token selector (ETH, USDC, HUNT, PLOT). ETH is payable, USDC/HUNT require approval to Zap, PLOT uses existing direct flow. USDC displayed with 6 decimals. Sell tab unchanged.

## Test plan
- [ ] Verify lint and typecheck pass
- [ ] Verify ETH/USDC/HUNT/PLOT toggle renders on buy tab
- [ ] Verify PLOT mode works (existing behavior)
- [ ] Test ETH mode quote estimation
- [ ] Test USDC mode approval + zap flow
- [ ] Verify sell tab unchanged
- [ ] Verify trade indexing fires for all modes
- [ ] Verify selector hidden when ZAP_PLOTLINK is zero address

Fixes realproject7/agent-os#244, Fixes realproject7/agent-os#245

### DIFF
--- a/lib/contracts/constants.ts
+++ b/lib/contracts/constants.ts
@@ -33,10 +33,10 @@ export const STORY_FACTORY = (process.env.NEXT_PUBLIC_CONTRACT_ADDRESS ??
     ? "0xfa5489b6710Ba2f8406b37fA8f8c3018e51FA229"
     : "0x337c5b96f03fB335b433291695A4171fd5dED8B0")) as `0x${string}`;
 
-/** ZapPlotLink — one-click buy (ETH -> PLOT -> storyline token via Uniswap V4 + MCV2) */
+/** ZapPlotLinkV2 — one-click buy (ETH/USDC/HUNT -> PLOT -> storyline token via Uniswap V4 + MCV2) */
 export const ZAP_PLOTLINK = (IS_TESTNET
   ? "0xC7C47D820D2D5b09797be2F438Cf329Ad7315682"
-  : "0x0000000000000000000000000000000000000000") as `0x${string}`;
+  : "0xEF6a8640c836b16Eb8cCD8016Ead4C8517aC3033") as `0x${string}`;
 
 /** $PLOT protocol token
  *  Testnet: PL_TEST ERC-20 on Base Sepolia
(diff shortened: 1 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The mainnet path looks directionally right, but this PR regresses the existing Sepolia/testnet frontend by wiring the new V2 ABI against the old V1 testnet zap address and by exposing mainnet USDC/HUNT addresses on all chains.

### Findings
- [high] `lib/zap.ts` now always uses the V2 ABI, but `ZAP_PLOTLINK` still points to the old Sepolia V1 contract on testnet. On Base Sepolia, `getZapQuote()` / `buildZapMintTx()` will call `estimateMint` / `mint(fromToken,...)` against `0xC7C47D...`, which is the previous ETH-only zap and does not match this ABI. That breaks the existing testnet zap flow instead of preserving it.
  - File: `lib/contracts/constants.ts:37`
  - Suggestion: either gate the V2 frontend to mainnet only (hide zap modes on testnet), or deploy/update a matching Sepolia V2 zap address before switching the frontend ABI globally.
- [high] `USDC`, `HUNT`, and `SUPPORTED_ZAP_TOKENS` are exported unconditionally with Base mainnet token addresses. Because the widget only checks whether `ZAP_PLOTLINK` is nonzero, Base Sepolia users will still see USDC/HUNT options and the app will check balances / approvals against mainnet token contracts on the wrong chain.
  - File: `lib/contracts/constants.ts:55`
  - Suggestion: make the supported zap token list chain-aware, or suppress non-PLOT zap options on testnet until Sepolia addresses and a matching zap deployment exist.

### Decision
Requesting changes because this PR introduces a cross-network frontend regression even if the mainnet integration itself is correct.

project7-interns (approved):
## T2b Review — APPROVED

No blockers. Core logic is solid across all three files.

### Verified correct
- ABI signatures match ZapPlotLinkV2 contract ✓
- `simulateContract` used for non-view estimates ✓
- ETH: payable with msg.value, no approval ✓
- USDC/HUNT: allowance check → approve → mint (two-step) ✓
- USDC 6 decimals handled correctly ✓
- 3% slippage buffer applied correctly ✓
- PLOT direct path unchanged ✓
- Sell tab unchanged ✓
- V4 Quoter calls fully removed ✓
- Trade indexing fires for all modes ✓
- isZapAvailable hides non-PLOT options when zap is zero ✓
- No dead code from v1 ✓
- Constants match spec ✓

### WARN (non-blocking)

1. **Button label when zap unavailable**: When `isZapAvailable` is false, `payToken` defaults to "ETH" but `isPlotMode` is true — button says "Buy with ETH" instead of "Buy with PLOT". Trade executes correctly, just label is misleading.

2. **Stale `useCallback` dep**: `refetchBalance` captures `isEthMode` but doesn't list it in dependencies. Could cause stale refetch when switching pay tokens.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The Sepolia gating fix resolves the cross-network regression, and the updated multi-token frontend wiring is consistent with the deployed V2 contract. CI is green.

### Findings
- No blocking findings.

### Decision
Approving because the mainnet multi-token zap flow is wired correctly, testnet is safely gated off, and the build is passing.

PR #527, opened 2026-03-25
[#521] Targeted notifications for token holders

### LINKED ISSUE #521: Targeted notifications for token holders — price alerts and new plot updates
## Summary

Send targeted notifications to users who hold specific storyline tokens, instead of broadcasting to all users.

### Two notification types
1. **New plot linked** — when a new plot is published, notify users who hold that storyline's tokens
2. **Price change alert** — when a storyline token price changes >10% up or down, notify holders

### Feasibility Assessment
Current structure broadcasts to all users via \`getEnabledTokens()\`. Targeted notifications require:

1. **FID → wallet mapping**: Already available — \`notification_tokens\` stores FID, and we can resolve FID → wallet via Farcaster/Neynar API. Alternatively, store wallet address alongside FID when saving notification tokens (from \`sdk.context.user\`).

2. **Token holder lookup**: Query MCV2_Bond trade_history for unique traders, then multicall \`balanceOf\` to filter current holders. This is the same pattern already used in the profile page (\`StoriesTab\` holder count logic).

3. **Price monitoring**: Add a cron that checks token prices periodically, compares to last known price, and queues notifications when >10% change detected. Needs a \`token_price_snapshots\` table.

### Implementation — reference Dropcast's targeted notification pattern
- Dropcast uses a \`notification_queue\` table with \`target_fid\` for per-user targeting
- A cron job processes pending notifications in batches
- Apply the same queue-based pattern to PlotLink

### Suggested approach
1. Add \`wallet_address\` column to \`notification_tokens\` table (populated from SDK context on save)
2. Create \`notification_queue\` table: \`id, target_fid, notification_type, storyline_id, title, body, target_url, status, created_at\`
3. Create \`token_price_snapshots\` table for price change tracking
4. Add \`queueHolderNotifications(storylineId, type, title, body)\` — finds holders, queues per-user notifications
5. Add price monitoring cron (every 5 min) that compares current vs last snapshot
6. Process queue via existing cron or new dedicated cron

### Acceptance Criteria
- [ ] New plot notifications sent only to storyline token holders
- [ ] Price change (>10%) alerts sent only to token holders
- [ ] Notification queue table for reliable delivery
- [ ] Price snapshot tracking for change detection

realproject7:
## Summary
- **DB migration** (`00022`): adds `wallet_address` to `notification_tokens`, creates `notification_queue` and `token_price_snapshots` tables
- **Targeted new plot notifications**: `notifyNewPlot()` now queries on-chain `balanceOf` to find storyline token holders, sends only to them (falls back to all users if no holders found)
- **Price change alerts**: `checkPriceChangeAlert()` compares current price to last snapshot, sends alerts to holders when >10% movement detected
- **Price alerts cron**: new `/api/cron/price-alerts` route iterates active storylines and checks for price changes
- **Wallet address capture**: save-token API now accepts `walletAddress` parameter to link FID → wallet

## Architecture
- Token holder targeting uses multicall `balanceOf` against all enabled notification tokens with wallet addresses
- Price snapshots stored in `token_price_snapshots` table, compared against previous snapshot each cron run
- `notification_queue` table created for future batched delivery (currently direct send, queue available for future optimization)
- Bumped version to `0.1.6`

## Test plan
- [ ] Migration applies cleanly (`wallet_address` column, new tables)
- [ ] New plot notification targets only token holders (verify with test wallet)
- [ ] Falls back to all users when no holders have notification tokens
- [ ] Price alert fires when token moves >10%
- [ ] `/api/cron/price-alerts` processes all active storylines
- [ ] save-token API accepts and stores `walletAddress`
- [ ] All 104 existing tests pass

Fixes #521

### DIFF
--- a/lib/notifications.server.ts
+++ b/lib/notifications.server.ts
@@ -13,6 +13,8 @@ import { createPublicClient, http, erc20Abi, type Address } from "viem";
 import { base } from "viem/chains";
 import { STORY_FACTORY } from "./contracts/constants";
 
+const NEYNAR_BASE = "https://api.neynar.com/v2/farcaster";
+
 const supabaseUrl = process.env.NEXT_PUBLIC_SUPABASE_URL || "";
 const supabaseServiceKey = process.env.SUPABASE_SERVICE_ROLE_KEY || "";
 
@@ -42,9 +44,11 @@ export async function saveUserNotificationToken(
   token: string,
   url: string,
   clientAppFid?: number,
-  walletAddress?: string,
 ): Promise<void> {
   const supabase = getSupabase();
+
+  // Resolve wallet from FID via trusted Neynar API
+  const walletAddress = await resolveWalletForFid(fid);
 
   const row: Record<string, unknown> = {
     fid,
@@ -55,7 +59,7 @@ export async function saveUserNotificationToken(
     updated_at: new Date().toISOString(),
   };
   if (walletAddress) {
-    row.wallet_address = walletAddress.toLowerCase();
+    row.wallet_address = walletAddress;
   }
 
   const { error } = await supabase
@@ -99,6 +103,34 @@ export async function getEnabledTokens(): Promise<NotificationToken[]> {
     notificationToken: row.notification_token,
     notificationUrl: row.notification_url,
   }));
+}
+
+// ---- [#521] FID → Wallet Resolution (trusted) ----
+
+/**
+ * Resolve a Farcaster FID to its verified Ethereum address via Neynar API.
+ * Returns the first verified address, or null if unavailable.
+ */
+async function resolveWalletForFid(fid: number): Promise<string | null> {
+  const apiKey = process.env.NEYNAR_API_KEY;
+  if (!apiKey) return null;
+
+  try {
+    const res = await fetch(`${NEYNAR_BASE}/user/bulk?fids=${fid}`, {
+      headers: { accept: "application/json", "x-api-key": apiKey },
+      signal: AbortSignal.timeout(3000),
+    });
+    if (!res.ok) return null;
+    const json = await res.json();
+    const user = json.users?.[0];
+    if (!user) return null;
+
+    // Use first verified Ethereum address
+    const verifiedAddress = user.verified_addresses?.eth_addresses?.[0];
+    return verifiedAddress?.toLowerCase() ?? null;
+  } catch {
+    return null;
+  }
 }
 
 // ---- [#521] Token Holder Targeting ----
@@ -264,19 +296,10 @@ export async function notifyNewPlot(
     .eq("contract_address", STORY_FACTORY.toLowerCase())
     .single();
 
-  let tokens: NotificationToken[];
+  // Only notify token holders — no broadcast fallback
+  if (!storyline?.token_address) return;
 
-  if (storyline?.token_address) {
-    // Target only token holders
-    tokens = await getTokenHolderTokens(storyline.token_address);
-    if (tokens.length === 0) {
-      // Fallback: send to all if no holders have notification tokens
-      tokens = await getEnabledTokens();
-    }
-  } else {
-    tokens = await getEnabledTokens();
-  }
-
+  const tokens = await getTokenHolderTokens(storyline.token_address);
   if (tokens.length === 0) return;
 
   await sendNotification({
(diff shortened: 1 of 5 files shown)


project7-interns (changes requested):
REQUEST CHANGES — two findings:

**1. Bug: `alerts` counter never incremented (price-alerts/route.ts:47)**
`const alerts = 0;` — declared as const and never incremented, but returned in the response. Should be `let alerts = 0;` and incremented after each successful `checkPriceChangeAlert()` call. Currently the API always returns `{ checked: N, alerts: 0 }` regardless of how many alerts fire.

**2. Missing wallet address validation (save-token/route.ts)**
`walletAddress` from the request body is passed through to the DB with no format validation. Since this is a system boundary (user input → DB), it should validate that the value is a valid Ethereum address (e.g. `/^0x[0-9a-fA-F]{40}$/`) before storage. Invalid values would cause silent failures in the downstream `balanceOf` multicall.

Otherwise solid: multicall token holder targeting with fallback is clean, migration is well-structured with proper indexes, price snapshot comparison logic is correct.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The notification targeting direction is right, but the current implementation still has one behavior that broadcasts beyond holders and one trust boundary issue in the holder-mapping flow.

### Findings
- [high] The new-plot path still broadcasts to all enabled users when no holder notification tokens are found, which violates the targeted-only acceptance criteria.
  - File: `lib/notifications.server.ts:272`
  - Suggestion: do not fall back to `getEnabledTokens()` for holder-targeted plot notifications; if no subscribed holders are found, skip delivery instead of broadcasting.
- [high] `walletAddress` is accepted directly from the client and persisted as the holder-mapping key without any verification, so a caller can register someone else's wallet and receive that wallet's holder-only alerts.
  - File: `src/app/api/notifications/save-token/route.ts:10`
  - Suggestion: derive the wallet from trusted Farcaster / miniapp context or another verified source instead of trusting raw request JSON.

### Decision
Requesting changes because issue #521 explicitly requires new-plot and price-change notifications to target token holders only. The current fallback-to-all path breaks that behavior, and the unverified `walletAddress` input makes the targeting list spoofable.

project7-interns (approved):
Re-reviewed — both T2b findings fixed:
1. alerts counter now uses let and increments correctly (checkPriceChangeAlert returns boolean)
2. walletAddress validated with /^0x[0-9a-fA-F]{40}$/ regex, invalid values silently dropped

APPROVE from T2b.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Re-checking against the live issue scope, the current implementation still misses two explicit targeting requirements.

### Findings
- [high] The new-plot notification path still broadcasts beyond holders.
  - File: `lib/notifications.server.ts:272`
  - Suggestion: remove the `getEnabledTokens()` fallback for holder-targeted plot notifications; if there are no subscribed holders, skip delivery instead of broadcasting.
- [high] The holder mapping still trusts client-provided wallet identity instead of a trusted context-derived mapping.
  - File: `src/app/api/notifications/save-token/route.ts:10`
  - Suggestion: derive the wallet from trusted Farcaster / miniapp context as described in issue #521, rather than persisting whatever wallet the client submits.

### Decision
Requesting changes because the live issue still says: targeted notifications should go to users who hold the storyline token instead of broadcasting to all users, and the suggested wallet mapping source is `sdk.context.user` / a trusted Farcaster mapping. The current fallback-to-all behavior and raw client wallet mapping still do not meet that scope.

project7-interns (approved):
Re-reviewed after T2a blocker fixes. Both addressed:
1. No broadcast fallback — notifyNewPlot returns early if no holders, no getEnabledTokens() fallback
2. Trusted wallet — resolved server-side via Neynar verified_addresses, walletAddress removed from save-token route

APPROVE from T2b.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The current head addresses the two in-scope blockers from my prior reviews. New-plot notifications are now holder-only with no broadcast fallback, and wallet mapping is resolved server-side from trusted Farcaster/Neynar data instead of client-supplied input.

### Findings
- None.

### Decision
Approving because the current implementation now matches issue #521's targeting requirements: holder-only delivery for new-plot and price alerts, plus trusted wallet resolution for FID-to-wallet mapping.

Issue #555, opened 2026-03-26
Persist language filter preference in localStorage


realproject7:
## Summary

When a user sets a language filter (e.g. Korean), it resets to the default on revisit. The filter preference should persist across sessions using localStorage.

### Implementation
- Save the selected language filter to `localStorage` (e.g. key: `plotlink-lang-filter`)
- On page load, read from localStorage and use as the initial filter value
- If no saved preference, default to "all"
- Apply the same pattern to writer filter and genre filter if they exist as persistent preferences

### Acceptance Criteria
- [ ] Language filter persists across page reloads and revisits
- [ ] Stored in localStorage (not cookies)
- [ ] Default is "all" when no saved preference exists
- [ ] Changing the filter updates localStorage immediately

realproject7:
Fixed via PR #557. Language filter now persists in localStorage across sessions.

PR #557, opened 2026-03-26
[#555] Persist language filter in localStorage

### LINKED ISSUE #275: [bug] Language 'All' filter still reverts to English — two-part fix needed
## Bug

Selecting "All languages" on the discovery page always reverts to English. PR #273 fixed part of this but PR #274 (custom dropdown) was based on an older branch and reverted the fix. Additionally, the filter component itself has a logic gap.

### 1. `page.tsx` fallback was reverted
PR #273 changed the lang fallback to allow `"all"`, but PR #274 overwrote it back to:
```typescript
const lang = rawLang && (LANGUAGES as readonly string[]).includes(rawLang) ? rawLang : "English";
```
`"all"` is not in `LANGUAGES`, so it's rejected → falls back to `"English"`.

**Fix**: Re-apply the PR #273 fix:
```typescript
const lang = rawLang === "all" ? "all" : rawLang && (LANGUAGES as readonly string[]).includes(rawLang) ? rawLang : "English";
```

### 2. `LanguageFilter` omits `lang` param when "all" is selected
In `src/components/GenreLanguageFilter.tsx`:
```typescript
if (value !== "all") params.set("lang", value);
```
When "All languages" is selected, `lang` is not included in the URL. The page then loads with no `lang` param → defaults to English.

**Fix**: Always include `lang` when it's explicitly `"all"`:
```typescript
if (value !== "all") params.set("lang", value);
else params.set("lang", "all");
```
Or simplify to always set it: `params.set("lang", value);`

## Files to modify
- `src/app/page.tsx` — line 36: allow `"all"` as valid lang value
- `src/components/GenreLanguageFilter.tsx` — always pass `lang=all` in URL when selected

## Acceptance criteria
- [ ] Selecting "All languages" shows stories in all languages
- [ ] Default (no URL param) still shows English
- [ ] Filter persists on page reload

realproject7:
## Summary
Fixes #275

Saves the user's language filter selection to localStorage and restores it on revisit.

- On language filter change, saves to `localStorage.plotlink_lang`
- On page load with no `lang` URL param (default "all"), reads saved preference and applies it via `router.replace`
- Validates saved value against `LANGUAGES` list before applying
- Falls back to "all" when no saved preference or invalid value

## Test plan
- [ ] Select "Korean" language filter, navigate away, return to home — should show Korean
- [ ] Select "All" language filter — should clear preference (shows all on return)
- [ ] Clear localStorage manually, visit home — should show all languages
- [ ] Direct link with `?lang=French` should override saved preference

### DIFF
--- a/src/components/FilterBar.tsx
+++ b/src/components/FilterBar.tsx
@@ -55,7 +55,8 @@ export function FilterBar({ writer, genre, lang, tab }: FilterBarProps) {
 
   // Restore saved language preference on first visit (no lang param in URL)
   useEffect(() => {
-    if (lang !== "all") return;
+    const urlParams = new URLSearchParams(window.location.search);
+    if (urlParams.has("lang")) return; // explicit param — don't override
     try {
       const saved = localStorage.getItem("plotlink_lang");
       if (saved && saved !== "all" && (LANGUAGES as readonly string[]).includes(saved)) {
--- a/src/components/FilterBar.tsx
+++ b/src/components/FilterBar.tsx
@@ -53,12 +53,24 @@ export function FilterBar({ writer, genre, lang, tab }: FilterBarProps) {
     return () => document.removeEventListener("mousedown", handleClick);
   }, []);
 
+  // Restore saved language preference on first visit (no lang param in URL)
+  useEffect(() => {
+    if (lang !== "all") return;
+    try {
+      const saved = localStorage.getItem("plotlink_lang");
+      if (saved && saved !== "all" && (LANGUAGES as readonly string[]).includes(saved)) {
+        router.replace(buildHref({ tab, writer, genre, lang: saved }));
+      }
+    } catch {}
+  }, []); // eslint-disable-line react-hooks/exhaustive-deps
+
   function toggle(key: FilterKey) {
     setOpen(open === key ? null : key);
   }
 
   function navigate(params: { tab: string; writer: string; genre: string; lang: string }) {
     setOpen(null);
+    try { localStorage.setItem("plotlink_lang", params.lang); } catch {}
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
LGTM. Clean localStorage persistence — URL params correctly take precedence over saved preference, saved values validated against LANGUAGES list, unnecessary redirects avoided when saved=all, and try/catch handles environments without localStorage.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The localStorage restore logic cannot distinguish a missing `lang` param from an explicit `?lang=all`, so it overrides valid explicit links with the saved preference.

### Findings
- [medium] `FilterBar` restores `plotlink_lang` whenever the normalized `lang` prop is `all`, but `src/app/page.tsx` already normalizes both a missing `lang` param and an explicit `?lang=all` to the same `lang="all"` prop. That means a user who intentionally opens `/?lang=all` will still be redirected to their saved language, which violates the ticket scope of restoring only when there is no `lang` URL param.
  - File: `src/components/FilterBar.tsx:58`
  - Suggestion: Pass a separate boolean or raw search-param signal down from `src/app/page.tsx` so the restore effect only runs when the URL truly omitted `lang`.

### Decision
Request changes until the restore logic can tell `no lang param` apart from an explicit `?lang=all`.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The restore effect now checks the raw URL for a `lang` param before applying the saved preference, so explicit `?lang=all` links are no longer overridden.

### Findings
- None.

### Decision
Approve. The requested behavior is covered. Residual risk: GitHub checks were still in progress at review time.

PR #559, opened 2026-03-26
[#276] Fix stale language filter via Next.js Data Cache bypass

realproject7:
## Summary
Fixes #276

Root cause: Supabase's PostgREST client uses `fetch()` internally, which Next.js intercepts and caches via its Data Cache. With `revalidate = 120`, filtered query results (e.g., `?lang=English`) served stale data after migration 00026 updated language values in the DB. Each unique filter combination creates a separate cache entry, so some entries may not have been revalidated yet.

Fix: Add `cache: 'no-store'` to the server-side Supabase client's global fetch configuration, ensuring all PostgREST queries bypass Next.js's Data Cache and always hit the database.

The page-level `revalidate = 120` still applies to the Full Route Cache (HTML), which is fine since each searchParams combination gets its own entry and revalidates after 2 minutes.

## Test plan
- [ ] Deploy to Vercel, visit `/?lang=English` — should only show English stories
- [ ] Switch to `/?lang=Korean` — should only show Korean stories
- [ ] Run migration on DB, verify changes appear immediately without waiting 2 min

### DIFF
--- a/lib/supabase.ts
+++ b/lib/supabase.ts
@@ -25,6 +25,9 @@ export function createServiceRoleClient(): SupabaseClient<Database> | null {
       autoRefreshToken: false,
       persistSession: false,
     },
+    global: {
+      fetch: (url, options) => fetch(url, { ...options, cache: "no-store" }),
+    },
   });
 }
 
--- a/lib/supabase.ts
+++ b/lib/supabase.ts
@@ -25,9 +25,6 @@ export function createServiceRoleClient(): SupabaseClient<Database> | null {
       autoRefreshToken: false,
       persistSession: false,
     },
-    global: {
-      fetch: (url, options) => fetch(url, { ...options, cache: "no-store" }),
-    },
   });
 }
 
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -4,9 +4,8 @@ import { getTrendingStorylines } from "../../lib/ranking";
 import { StoryGrid } from "../components/StoryGrid";
 import { FilterBar, type WriterFilterValue } from "../components/FilterBar";
 import { GENRES, LANGUAGES } from "../../lib/genres";
+import { connection } from "next/server";
 import Link from "next/link";
-
-export const revalidate = 120;
 
 const TABS = ["new", "trending"] as const;
 type Tab = (typeof TABS)[number];
@@ -22,6 +21,9 @@ export default async function Home({
 }: {
   searchParams: SearchParams;
 }) {
+  // Opt out of Next.js Data Cache so filtered queries always hit the DB
+  await connection();
+
   const { tab: rawTab, writer: rawWriter, page: rawPage, genre: rawGenre, lang: rawLang } = await searchParams;
   const tab: Tab = TABS.includes(rawTab as Tab) ? (rawTab as Tab) : "new";
   const writer: WriterFilterValue = WRITER_VALUES.includes(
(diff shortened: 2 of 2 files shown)


project7-interns (approved):
LGTM. Verified that createServerClient() delegates to createServiceRoleClient(), so the cache: 'no-store' fix correctly applies to homepage queries. Root cause analysis is sound — Next.js Data Cache intercepting Supabase's internal fetch() calls. The page-level revalidate still handles HTML caching appropriately.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The proposed cache bypass is too broad. It fixes the filtered home-page read by making every service-role Supabase fetch uncached, which changes caching behavior across much more of the app than this ticket scoped.

### Findings
- [high] `cache: 'no-store'` is being added to the shared service-role client factory, so every `createServerClient()` consumer inherits uncached fetches. In Next.js, an uncached fetch opts the route out of the Full Route Cache, so this does not just fix the live language-filter query. It also makes pages like the home feed, story pages, OG routes, and any other server-side reads that use `createServerClient()` render dynamically on every request, defeating the existing `revalidate` behavior and creating a broad perf regression.
  - File: `lib/supabase.ts:28`
  - Suggestion: Scope the cache bypass to the specific read path that was observed stale on the live site, or invalidate/revalidate the affected route/data after the migration, rather than changing the shared client factory for the whole app.

### Decision
Request changes until the cache bypass is narrowed so it does not disable route caching for every server-side Supabase read.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
This narrows the change to the homepage, but it still does not appear to change the caching semantics for the affected query path.

### Findings
- [high] `await connection()` is being added before the Supabase query, but this page was already reading the request-time `searchParams` prop before issuing that query. In Next.js, fetches after request-time APIs are already treated as uncached by default, so this change is likely redundant and may not actually fix the stale-English-filter bug seen on the live site.
  - File: `src/app/page.tsx:21`
  - Suggestion: Either show a concrete repro proving `connection()` changes the live behavior here, or target the actual stale layer directly (for example by making the relevant data fetch explicit about cache behavior instead of changing route rendering semantics).

### Decision
Request changes until there is evidence this change affects the stale query path, or the fix is moved to the specific data-fetch layer causing the bug.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
This narrows the diff to a single line, but the proposed fix still does not appear to affect the stale layer for this route.

### Findings
- [high] Removing `export const revalidate = 120` from the homepage is likely a no-op here. This page already uses the Page `searchParams` prop, which Next.js treats as a Dynamic API that opts the route into request-time rendering. Next.js also does not cache fetches discovered after a Dynamic API is used. Given that the Supabase query runs after `await searchParams`, the stale-English-filter bug is unlikely to be caused by this route-level `revalidate` export.
  - File: `src/app/page.tsx:7`
  - Suggestion: Show a concrete live repro proving that deleting `revalidate` changes the behavior on this page, or move the fix to the actual stale layer instead of changing a route setting that should already be bypassed by the page's dynamic `searchParams` usage.

### Decision
Request changes until the fix targets a cache layer that is demonstrably involved in the stale live-site behavior.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The branch is now minimal, but I still do not see evidence that deleting the homepage `revalidate` export changes the stale layer involved in this bug.

### Findings
- [high] `src/app/page.tsx` already uses the Page `searchParams` prop, which Next.js documents as a Dynamic API that opts the page into request-time rendering. Next.js also documents that `fetch` requests discovered after a Dynamic API are not cached by default. Since the Supabase read happens after `await searchParams`, removing `export const revalidate = 120` still looks unlikely to affect the stale English-filter behavior on the live site.
  - File: `src/app/page.tsx:7`
  - Suggestion: Show a concrete repro that this line removal changes the live behavior, or move the fix to the specific cache/data layer that is demonstrably serving stale results.

### Decision
Request changes until the fix targets a cache layer that is actually implicated by the live-site bug.

PR #568, opened 2026-03-26
[#567] Support non-Farcaster wallet users

### LINKED ISSUE #567: Support non-Farcaster wallet users — make FID nullable, fix register flow
## Summary

PlotLink must work for ALL wallet users, not just Farcaster users. Currently, connecting a wallet without a Farcaster account fails with "No Farcaster account found for this wallet" (HTTP 404). Non-FID users should be first-class citizens with NULL for Farcaster-specific fields.

**This is critical — PlotLink is a writing platform for everyone, not a Farcaster-only app.**

#### 1. Schema: Make `fid` nullable
\`supabase/migrations/00027_create_users_table.sql\` has \`fid INTEGER UNIQUE NOT NULL\`.

Fix: \`fid INTEGER UNIQUE\` (drop NOT NULL). Create a new migration:
\`\`\`sql
ALTER TABLE users ALTER COLUMN fid DROP NOT NULL;
\`\`\`

Non-FID users need a different unique identifier. Options:
- Add \`wallet_address TEXT UNIQUE\` as an alternative key (the primary connected wallet)
- Use \`primary_address\` as the unique key for non-FID users
- Composite: unique on either \`fid\` (when present) or \`primary_address\` (when fid is null)

**Recommended**: Add a unique index on \`primary_address\` since every user has a wallet but not every user has a FID.

#### 2. Register flow: Don't reject non-Farcaster wallets
\`src/app/api/user/register-by-wallet/route.ts\` lines 58-66 returns 404 if no Farcaster account found.

Fix: When both SteemHunt and Neynar return nothing:
- Create a user record with \`fid = NULL\`
- Set \`primary_address\` and \`verified_addresses\` from the connected wallet
- All Farcaster fields (username, display_name, pfp_url, bio, etc.) remain NULL
- The user can still create stories, trade, comment (with wallet signature), rate, donate
- Profile page shows truncated address instead of Farcaster identity (this already works)

#### 3. User data builder: Handle no FID
\`lib/user-data.ts\` lines 53 and 74 always set \`fid\` from SteemHunt or Neynar.

Fix: Add a third path when both sources return null:
\`\`\`typescript
// Wallet-only user (no Farcaster)
return {
  ...base,
  fid: null,
  primary_address: verifiedAddresses[0],
  // All Farcaster fields remain null/default
} as UserInsert;
\`\`\`

#### 4. Upsert logic: Handle null FID
\`register-by-wallet/route.ts\` line 115 uses \`.eq("fid", userData.fid)\` for conflict resolution.

Fix: For non-FID users, use \`primary_address\` as the upsert key instead:
- If \`fid\` is present: upsert by \`fid\` (existing behavior)
- If \`fid\` is null: upsert by \`primary_address\`

#### 5. Existing user lookup: Also search by wallet
\`register-by-wallet/route.ts\` line 35-39 checks \`verified_addresses\`.

Fix: Also check \`primary_address\` for wallet-only users who don't have \`verified_addresses\` populated from Farcaster.

#### 6. Profile page: Handle no FID gracefully
\`src/app/profile/[address]/page.tsx\` should already work since it uses wallet address as the route param, but verify:
- Profile header shows wallet address when no Farcaster identity
- No errors when user record has null FID
- Writer stats, portfolio, activity all work with wallet-only users

#### 7. All user-facing features must work without FID
Verify these work for wallet-only users:
- [ ] Create storyline
- [ ] Chain plot
- [ ] Trade (buy/sell via Zap)
- [ ] Comment (wallet signature based — no FID needed)
- [ ] Rate stories
- [ ] Donate
- [ ] View profile
- [ ] Activity feed

### What NOT to break
- FID users must still work exactly as before
- SteemHunt → Neynar fallback chain unchanged for FID users
- Quotient score only fetched when FID is available (already guarded)
- X stats only fetched when twitter handle is available (already guarded)

### Files to Modify
- New: \`supabase/migrations/00028_users_fid_nullable.sql\`
- Modify: \`src/app/api/user/register-by-wallet/route.ts\`
- Modify: \`lib/user-data.ts\`
- Modify: \`src/app/profile/[address]/page.tsx\` (verify no-FID rendering)
- Modify: \`src/app/api/user/onboard/route.ts\` (verify no-FID path)

### Acceptance Criteria
- [ ] Non-Farcaster wallet users can connect without errors
- [ ] User record created with \`fid = NULL\`, wallet address stored
- [ ] All PlotLink features work for wallet-only users (create, trade, comment, rate, donate)
- [ ] FID users unaffected — same behavior as before
- [ ] Profile page renders correctly for wallet-only users (truncated address, no Farcaster data)
- [ ] No 404 or 500 errors on wallet connect for non-Farcaster wallets

realproject7:
## Summary
- **Migration 00028**: make `fid` nullable, add unique index on `primary_address` for wallet-only user upserts
- **register-by-wallet**: no longer returns 404 for non-Farcaster wallets — creates user record with `fid=NULL`, wallet address stored as `primary_address`
- **buildUserData**: third path for wallet-only users (no SteemHunt, no Neynar)
- **Upsert logic**: uses `primary_address` as key when `fid` is null (both register-by-wallet and onboard routes)
- **User lookup**: `getUserFromDB` and all route lookups now search by `primary_address` as fallback
- **getFarcasterProfile**: skips DB shortcut for wallet-only users so they don't get a broken profile object
- **Supabase types**: `fid` is now `number | null` across Row/Insert/Update
- All existing FID user flows are completely unaffected

Fixes #567

## Test plan
- [ ] Connect a non-Farcaster wallet → verify user record created (fid=NULL)
- [ ] Profile page renders correctly for wallet-only user (truncated address, no FC data)
- [ ] Connect a Farcaster wallet → verify existing behavior unchanged
- [ ] Run migration 00028 against Supabase
- [ ] Verify build passes (`npm run build`)

### DIFF
--- a/lib/actions.ts
+++ b/lib/actions.ts
@@ -15,9 +15,9 @@ import type { Address } from "viem";
 export async function getFarcasterProfile(
   address: string,
 ): Promise<FarcasterProfile | null> {
-  // Try DB first
+  // Try DB first (only if user has a FID — wallet-only users have no Farcaster profile)
   const dbUser = await getUserFromDB(address);
-  if (dbUser) {
+  if (dbUser && dbUser.fid != null) {
     return {
       fid: dbUser.fid,
       username: dbUser.username ?? "",
@@ -41,6 +41,7 @@ export async function fetchAgentMetadata(
 
 /**
  * Look up a user from the DB by wallet address.
+ * Searches verified_addresses first, then primary_address for wallet-only users.
  */
 export async function getUserFromDB(
   address: string,
@@ -50,11 +51,19 @@ export async function getUserFromDB(
 
   const normalized = address.toLowerCase();
 
-  const { data } = await supabase
+  const { data: byVerified } = await supabase
     .from("users")
     .select("*")
     .contains("verified_addresses", [normalized])
     .single();
 
-  return data ?? null;
+  if (byVerified) return byVerified;
+
+  const { data: byPrimary } = await supabase
+    .from("users")
+    .select("*")
+    .eq("primary_address", normalized)
+    .single();
+
+  return byPrimary ?? null;
 }
--- a/lib/supabase.ts
+++ b/lib/supabase.ts
@@ -394,7 +394,7 @@ export interface Database {
       users: {
       Row: {
         id: string;
-        fid: number;
+        fid: number | null;
         username: string | null;
         display_name: string | null;
         pfp_url: string | null;
@@ -429,7 +429,7 @@ export interface Database {
       };
       Insert: {
         id?: never;
-        fid: number;
+        fid?: number | null;
         username?: string | null;
         display_name?: string | null;
         pfp_url?: string | null;
@@ -464,7 +464,7 @@ export interface Database {
       };
       Update: {
         id?: never;
-        fid?: number;
+        fid?: number | null;
         username?: string | null;
         display_name?: string | null;
         pfp_url?: string | null;
--- a/lib/user-data.ts
+++ b/lib/user-data.ts
@@ -69,12 +69,21 @@ export async function buildUserData(opts: {
     } as UserInsert;
   }
 
+  if (neynarProfile) {
+    return {
+      ...base,
+      fid: neynarProfile.fid,
+      username: neynarProfile.username,
+      display_name: neynarProfile.displayName,
+      pfp_url: neynarProfile.pfpUrl,
+      bio: neynarProfile.bio,
+    } as UserInsert;
+  }
+
+  // Wallet-only user (no Farcaster account)
   return {
     ...base,
-    fid: neynarProfile!.fid,
-    username: neynarProfile!.username,
-    display_name: neynarProfile!.displayName,
-    pfp_url: neynarProfile!.pfpUrl,
-    bio: neynarProfile!.bio,
+    fid: null,
+    primary_address: verifiedAddresses[0] || null,
   } as UserInsert;
 }
--- a/src/app/api/user/onboard/route.ts
+++ b/src/app/api/user/onboard/route.ts
@@ -32,12 +32,23 @@ export async function POST(request: NextRequest) {
       );
     }
 
-    // Check existing user and cooldown
-    const { data: existingUser } = await supabase
+    // Check existing user (by verified_addresses or primary_address)
+    let existingUser = null;
+    const { data: byVerified } = await supabase
       .from("users")
       .select("*")
       .contains("verified_addresses", [normalizedAddress])
       .single();
+    if (byVerified) {
+      existingUser = byVerified;
+    } else {
+      const { data: byPrimary } = await supabase
+        .from("users")
+        .select("*")
+        .eq("primary_address", normalizedAddress)
+        .single();
+      existingUser = byPrimary;
+    }
 
     // Enforce 5-min cooldown on ALL refreshes
     if (existingUser?.steemhunt_fetched_at) {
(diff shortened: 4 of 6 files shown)


project7-interns (approved):
T2b approves. Migration correctly handles nullable fid with partial unique index on primary_address. Dual lookup pattern is consistent across all routes. Wallet-only user path in buildUserData is clean. No issues found.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The wallet-only creation path is in place, but the new upsert logic still breaks when an existing wallet-only user later gains a Farcaster identity.

### Findings
- [high] Both update paths choose the lookup key from the new payload (`userData.fid`) instead of from the row that already exists. If a wallet-only row already exists for `primary_address = <wallet>` and that wallet later resolves to a Farcaster account, the insert will conflict on the new unique `primary_address` index, then the follow-up update will target `fid = <new fid>` even though the existing row still has `fid = NULL`. That means the update matches no row and the route 500s instead of upgrading the wallet-only user into an FID-backed user.
  - File: `src/app/api/user/register-by-wallet/route.ts:114`
  - File: `src/app/api/user/onboard/route.ts:115`
  - Suggestion: when `existingUser` was found, update by that existing row's identity (for example `id` or `primary_address`) rather than by `userData.fid`. Otherwise the wallet-only -> Farcaster transition is broken.

### Decision
Requesting changes because the current upsert logic fails a realistic account-upgrade path and does not fully keep wallet-only and FID users interoperable.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The wallet-only -> Farcaster upgrade path is fixed, but the latest fallback still regresses a normal FID-user refresh path.

### Findings
- [high] In `register-by-wallet`, the unique-violation fallback now updates by `primary_address` when `existingUser` was not found at lookup time. That breaks Farcaster users who connect through a newly verified wallet that is not yet present in the cached DB row. In that case the initial lookup misses, the insert conflicts on the existing unique `fid`, and the fallback update targets `primary_address = <new wallet>` even though the existing FID row usually has a different `primary_address`. The update matches no row and the route 500s, which violates the ticket's requirement that FID users remain unaffected.
  - File: `src/app/api/user/register-by-wallet/route.ts:116`
  - Suggestion: on unique-violation fallback, update by the actual conflicting identity. For FID-backed payloads that should still be `fid = userData.fid`; reserve the `primary_address` fallback for the wallet-only (`fid = NULL`) case or query the conflicting row first.

### Decision
Requesting changes because the current fallback can still break existing Farcaster users when they connect through an address that has not yet been cached locally.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
`register-by-wallet` now handles both conflict cases, but `onboard` still has the same stale-wallet / existing-FID conflict gap.

### Findings
- [high] `src/app/api/user/onboard/route.ts` still only inserts when `existingUser` lookup misses, with no unique-violation recovery path. If an existing Farcaster user connects through a newly verified wallet that is not yet present in the cached DB row, the lookup misses, the insert conflicts on the existing unique `fid`, and the route returns 500 instead of refreshing the existing user. That means FID users are still not fully unaffected by this change.
  - File: `src/app/api/user/onboard/route.ts:131`
  - Suggestion: mirror the `register-by-wallet` conflict handling here as well: if insert hits `23505`, update by `fid` for FID-backed payloads and by `primary_address` for wallet-only payloads, or otherwise resolve and update the conflicting row.

### Decision
Requesting changes because the manual refresh path still breaks a realistic existing-user flow.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The remaining conflict-handling gaps are fixed. Both `register-by-wallet` and `onboard` now recover from unique-violation inserts by updating the conflicting row through the correct identity, so wallet-only users and existing FID users both keep working.

### Findings
- None.

### Decision
Approving because the nullable-FID schema change, wallet-only user path, and conflict recovery logic now satisfy the Batch 25 / #567 requirements without regressing existing Farcaster users.

PR #572, opened 2026-03-26
[#564] Auto-save draft content on /create page

### LINKED ISSUE #564: Auto-save draft content on /create page to prevent accidental data loss
## Summary

Writers can lose their entire story if they accidentally navigate away, close the tab, or refresh the /create page. For a writing platform, this is critical to prevent.

### Implementation

**Auto-save to localStorage:**
- Save title, content, genre, language, and hasDeadline to localStorage on every keystroke (debounced ~1 second)
- Key: `plotlink_draft_create` (or per-user key if wallet connected)
- On page load, restore from localStorage if a draft exists
- Show a small "Draft restored" indicator when content is recovered

**Unsaved changes warning:**
- Add `beforeunload` event listener when content is non-empty
- Browser shows native "Leave site? Changes you made may not be saved" dialog
- Remove listener after successful publish

**Clear draft after publish:**
- Delete localStorage draft after storyline is successfully created
- Also clear when user explicitly clicks a "Discard draft" button

**Same pattern for Chain Plot page:**
- Auto-save draft plot content when writing a new chapter
- Key: `plotlink_draft_plot_{storylineId}`
- Clear after successful plot submission

### Acceptance Criteria
- [ ] Content auto-saved to localStorage on /create page (debounced)
- [ ] Draft restored on page revisit with indicator
- [ ] Browser warns before navigating away with unsaved content
- [ ] Draft cleared after successful publish
- [ ] "Discard draft" option available
- [ ] Same auto-save pattern on chain plot page
- [ ] Works across page refreshes and browser restarts

realproject7:
## Summary
- **useDraft hook** — generic localStorage auto-save with 1-second debounce, restore on mount, beforeunload warning, clear/discard helpers
- **Create Storyline form** — auto-saves title, content, genre, language to `plotlink_draft_create`
- **Chain Plot form** — auto-saves title, content to `plotlink_draft_plot_{storylineId}`
- **"Draft restored" banner** — shown for 3 seconds when content is recovered on page revisit
- **"Discard draft" button** — clears saved content and resets form fields
- **Auto-clear on publish** — drafts cleared via useEffect when state reaches "published"
- **beforeunload warning** — browser prompts before navigating away with unsaved content

Fixes #564

## Test plan
- [ ] Write content on /create, refresh page → draft restored with indicator
- [ ] Write content, close tab, reopen → draft restored
- [ ] Publish storyline → draft cleared from localStorage
- [ ] Click "Discard draft" → form cleared, localStorage cleared
- [ ] Try to navigate away with content → browser warns
- [ ] Chain plot draft saves per-storyline key
- [ ] Build passes (`npm run build`)

### DIFF
--- a/src/app/create/page.tsx
+++ b/src/app/create/page.tsx
@@ -200,6 +200,12 @@ function CreatePage() {
     discardDraft: discardChainDraft,
   } = useDraft(chainDraftKey, chainDraftValues, chainDraftSetters);
 
+  // Clear drafts on successful publish (must be above early returns — Rules of Hooks)
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => { if (newState === "published") clearNewDraft(); }, [newState]);
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  useEffect(() => { if (chainState === "published") clearChainDraft(); }, [chainState]);
+
   if (!isConnected) {
     return (
       <div className="flex min-h-[calc(100vh-2.75rem)] flex-col items-center justify-center gap-4 px-6">
@@ -277,12 +283,6 @@ function CreatePage() {
       </div>
     );
   }
-
-  // Clear drafts on successful publish
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  useEffect(() => { if (newState === "published") clearNewDraft(); }, [newState]);
-  // eslint-disable-next-line react-hooks/exhaustive-deps
-  useEffect(() => { if (chainState === "published") clearChainDraft(); }, [chainState]);
 
   const noStoryline = chainStorylineId === null;
 
--- a/src/app/create/page.tsx
+++ b/src/app/create/page.tsx
@@ -1,8 +1,9 @@
 "use client";
 
-import { Suspense, useState } from "react";
+import { Suspense, useState, useMemo, useEffect } from "react";
 import { useAccount } from "wagmi";
 import { useSearchParams } from "next/navigation";
+import { useDraft } from "../../hooks/useDraft";
 import { useQuery } from "@tanstack/react-query";
 import {
   validateContentLength,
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
REQUEST CHANGES — Rules of Hooks violation.

`src/app/create/page.tsx:278-281`: The two `useEffect` calls for clearing drafts on publish are placed AFTER the early return at `if (!isConnected) return`. React hooks must be called in the same order on every render. When `isConnected` changes from false to true, React will see additional hooks and throw an error. Move these useEffects above the early return, alongside the other hooks.

Minor (non-blocking): `discardDraft` in useDraft.ts eslint-disables the exhaustive-deps rule and omits `currentValues` — since it only checks `typeof` for resetting, this works but is fragile.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The autosave hook covers the basic save/restore flow, but the current chain-plot implementation still leaks one storyline's draft into another storyline's storage key.

### Findings
- [high] When the chain plot `storylineId` changes, `useDraft` re-runs with a new `key`, but if that new key has no saved draft it just returns early without clearing the previous form state. The old storyline's `title` / `content` remain in React state, and the debounced save then writes those stale values into the new storyline-specific key. That breaks the requirement that chain-plot drafts be isolated per storyline.
  - File: `src/app/create/page.tsx:186`
  - File: `src/hooks/useDraft.ts:24`
  - File: `src/hooks/useDraft.ts:63`
  - Suggestion: when the draft key changes and there is no saved draft for that key, reset the bound fields to their empty/default values before the autosave effect can persist the previous storyline's content under the new key.

### Decision
Requesting changes because the current implementation can overwrite one storyline's draft with another storyline's content, so the storyline-specific draft behavior is not correct yet.

project7-interns (approved):
T2b approves. Rules of Hooks violation fixed — useEffect calls now correctly placed before early return. useDraft hook is clean with proper debounce, restore, and beforeunload handling.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The remaining autosave issues are fixed. The draft hook now resets bound fields when the storyline-specific key changes and no draft exists for the new key, so chain-plot drafts no longer bleed across storylines, and the Rules-of-Hooks issue is already addressed.

### Findings
- None.

### Decision
Approving because the debounced localStorage autosave, restore/discard flow, beforeunload warning, and storyline-specific draft isolation now satisfy the Batch 26 / #564 requirements.

PR #601, opened 2026-03-27
Skip RPC fallback for known non-agent users

### LINKED ISSUE #600: Fix unnecessary RPC calls on /agents page for known non-agent users
## Bug

When a known PlotLink user (who is NOT an agent) visits `/agents`, the page fires 3+ RPC calls to `base-rpc.publicnode.com` (`agentIdByWallet`, `balanceOf`, `tokenOfOwnerByIndex`) even though the DB already knows this user has no agent registration. This defeats the purpose of PR #597 (agent DB cache).

Visible in network tab: multiple `base-rpc.publicnode.com` fetches on every page load.

## Root Cause

In `src/app/agents/page.tsx`, the detection flow is:

1. Call `getAgentUserFromDB(address)` — this function (lib/actions.ts lines 186-231) queries the users table with `agent_id IS NOT NULL` filter on ALL queries
2. For a user who exists in DB but has `agent_id = NULL` → all queries exclude them → returns `null`
3. `dbUser = null`, `dbDetected = false`
4. `needsRpcFallback = !dbLoading && !dbDetected && !!address` → **true**
5. Three RPC calls fire unnecessarily

**The function cannot distinguish between:**
- "User exists in DB but is not an agent" (`agent_id IS NULL`) → should NOT make RPC calls
- "User has no DB record at all" (completely unknown wallet) → SHOULD make RPC calls (might be externally registered agent)

## Fix

Change the detection to also check if the user exists in DB at all (regardless of agent status):

### Option A: Two-step DB check in agents/page.tsx

```typescript
// Step 1: Check if user has cached agent data
const { data: agentUser, isLoading: agentLoading } = useQuery({
  queryKey: ["db-agent", address],
  queryFn: () => getAgentUserFromDB(address!),
  enabled: !!address,
});

// Step 2: Check if user exists in DB at all (regardless of agent status)
const { data: regularUser, isLoading: userLoading } = useQuery({
  queryKey: ["db-user-exists", address],
  queryFn: () => getUserFromDB(address!),
  enabled: !!address && !agentLoading && agentUser === null,
});

const dbLoading = agentLoading || (agentUser === null && userLoading);
const dbDetected = agentUser?.agent_id != null;
const userExistsInDB = agentUser != null || regularUser != null;

// Only RPC if user has NO DB record at all (completely unknown wallet)
const needsRpcFallback = !dbLoading && !dbDetected && !userExistsInDB && !!address;
```

### Option B: Single function that returns discriminated result

Create a new `checkAgentStatus(address)` server action that returns:
```typescript
type AgentStatus = 
  | { status: "agent"; user: User }        // known agent
  | { status: "not-agent" }                 // known user, no agent_id  
  | { status: "unknown" }                   // no DB record at all
```

This is cleaner and avoids two separate DB round-trips.

## Expected behavior after fix

| Scenario | DB check | RPC calls | Result |
|----------|----------|-----------|--------|
| Known user, no agent_id | `getUserFromDB` → user found, `agent_id = NULL` | **None** | Show registration form |
| Known agent | `getAgentUserFromDB` → agent found | **None** | Show management UI |
| Unknown wallet (no DB record) | Both return null | Yes (3 calls) | Check if externally registered |
| Unknown wallet, externally registered agent | Both return null, RPC finds agent | Yes, then auto-cache | Show management UI |

## Files to modify

- `src/app/agents/page.tsx` — fix `needsRpcFallback` logic
- `lib/actions.ts` — possibly add `checkAgentStatus()` helper

## Branch

`task/600-fix-agent-rpc-fallback`

## Acceptance criteria

- [ ] Known PlotLink users with no agent_id → zero RPC calls on /agents page
- [ ] Known agents → zero RPC calls (read from DB)
- [ ] Completely unknown wallets → RPC fallback fires (correct)
- [ ] Externally registered agents detected and auto-cached on first visit
- [ ] Build passes

realproject7:
## Summary
Fixes #600

- Add `checkUserExists()` helper that checks if any DB row exists for a wallet (regardless of `agent_id`)
- Agents page and dashboard now distinguish:
  - **Known agent** (DB row with `agent_id != NULL`) → use DB, zero RPC calls
  - **Known non-agent** (DB row with `agent_id = NULL`) → skip RPC, show registration form
  - **Unknown wallet** (no DB record) → RPC fallback + auto-cache
- Eliminates 3+ unnecessary `agentIdByWallet`/`balanceOf`/`tokenOfOwnerByIndex` RPC calls per visit for the majority of users who are known non-agents

## Test plan
- [ ] Known non-agent user visits /agents → no RPC calls, sees Register tab immediately
- [ ] Known agent visits /agents → loaded from DB, sees Manage tab
- [ ] Unknown wallet visits /agents → RPC fallback fires, auto-caches result
- [ ] Build passes

### DIFF
--- a/lib/actions.ts
+++ b/lib/actions.ts
@@ -84,6 +84,17 @@ export async function fetchAgentMetadata(
 }
 
 /**
+ * Check if a user exists in the DB (any row, with or without agent_id).
+ * Returns true if a known user, false if completely unknown wallet.
+ */
+export async function checkUserExists(
+  address: string,
+): Promise<boolean> {
+  const user = await getUserFromDB(address);
+  return user !== null;
+}
+
+/**
  * Cache an externally registered agent by agentId.
  * Use when the wallet is an NFT owner (not the bound agent wallet),
  * so agentIdByWallet() wouldn't find it.
--- a/lib/contracts/erc8004.ts
+++ b/lib/contracts/erc8004.ts
@@ -251,7 +251,7 @@ export async function getAgentMetadata(
     });
     if (agentId <= BigInt(0)) return null;
 
-    const [uri, owner] = await Promise.all([
+    const [uri, owner, agentWallet] = await Promise.all([
       publicClient.readContract({
         address: ERC8004_REGISTRY,
         abi: erc8004Abi,
@@ -264,6 +264,12 @@ export async function getAgentMetadata(
         functionName: "ownerOf",
         args: [agentId],
       }).catch(() => undefined),
+      publicClient.readContract({
+        address: ERC8004_REGISTRY,
+        abi: erc8004Abi,
+        functionName: "getAgentWallet",
+        args: [agentId],
+      }).catch(() => undefined),
     ]);
     if (!uri) return null;
 
@@ -271,6 +277,7 @@ export async function getAgentMetadata(
     return {
       agentId: agentId.toString(),
       owner: owner as string | undefined,
+      agentWallet: agentWallet as string | undefined,
       name: (parsed.name as string) || "Unknown Agent",
       description: (parsed.description as string) || "",
       genre: (parsed.genre as string) || undefined,
--- a/src/app/agents/page.tsx
+++ b/src/app/agents/page.tsx
@@ -37,29 +37,9 @@ export default function AgentsPage() {
     enabled: !!address && !dbLoading && !dbDetected,
   });
 
-  const knownNonAgent = !dbDetected && !userExistsLoading && userExists === true;
-
-  // Lightweight background check for known non-agents: single agentIdByWallet call
-  // Catches users who registered externally after creating their PlotLink account
-  const { data: bgAgentId } = useReadContract({
-    address: ERC8004_REGISTRY,
-    abi: erc8004Abi,
-    functionName: "agentIdByWallet",
-    args: address ? [address] : undefined,
-    query: { enabled: knownNonAgent },
-  });
-
-  // If background check finds an agent, cache it
-  const bgFoundAgent = bgAgentId !== undefined && bgAgentId > BigInt(0);
-  const bgCachedRef = useRef(false);
-  useEffect(() => {
-    if (bgFoundAgent && address && !bgCachedRef.current) {
-      bgCachedRef.current = true;
-      cacheAgentById(address, bgAgentId!.toString()).catch(() => {});
-    }
-  }, [bgFoundAgent, address, bgAgentId]);
-
-  // Full RPC fallback: only for completely unknown wallets (no DB record at all)
+  // RPC fallback: only for completely unknown wallets (no DB record at all)
+  // Known users with agent_id=NULL are definitively non-agents — zero RPC calls
+  // External registrations are detected via profile refresh (/api/user/onboard)
   const needsRpcFallback = !dbLoading && !dbDetected && !userExistsLoading && userExists === false && !!address;
 
   const { data: rpcAgentId, isLoading: rpcWalletLoading } = useReadContract({
@@ -97,9 +77,6 @@ export default function AgentsPage() {
   if (dbDetected) {
     detectedAgentId = BigInt(dbAgentId!);
     detectedRole = dbIsOwner ? "owner" : dbIsAgentWallet ? "agentWallet" : "owner";
-  } else if (bgFoundAgent) {
-    detectedAgentId = bgAgentId;
-    detectedRole = "agentWallet";
   } else if (rpcIsOwner) {
     detectedAgentId = rpcOwnedToken;
     detectedRole = "owner";
--- a/src/app/agents/page.tsx
+++ b/src/app/agents/page.tsx
@@ -37,8 +37,29 @@ export default function AgentsPage() {
     enabled: !!address && !dbLoading && !dbDetected,
   });
 
-  // RPC fallback: only for completely unknown wallets (no DB record at all)
-  // Known users with agent_id=NULL are definitively non-agents — skip RPC
+  const knownNonAgent = !dbDetected && !userExistsLoading && userExists === true;
+
+  // Lightweight background check for known non-agents: single agentIdByWallet call
+  // Catches users who registered externally after creating their PlotLink account
+  const { data: bgAgentId } = useReadContract({
+    address: ERC8004_REGISTRY,
+    abi: erc8004Abi,
+    functionName: "agentIdByWallet",
+    args: address ? [address] : undefined,
+    query: { enabled: knownNonAgent },
+  });
+
+  // If background check finds an agent, cache it
+  const bgFoundAgent = bgAgentId !== undefined && bgAgentId > BigInt(0);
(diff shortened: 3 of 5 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new DB-existence check removes the unnecessary RPC calls for known non-agent users, but it also suppresses the external-agent fallback for any wallet that already has a PlotLink user row.

### Findings
- [high] Existing PlotLink users who later register an ERC-8004 agent in another app will now be permanently treated as non-agents, because `checkUserExists()` is used as a definitive skip-RPC signal even when the local DB cache is stale.
  - File: `src/app/agents/page.tsx:33`
  - File: `src/components/AgentDashboard.tsx:27`
  - File: `lib/actions.ts:90`
  - Suggestion: preserve the no-RPC path for users the app can prove are still non-agents, but do not treat any pre-existing user row as authoritative enough to disable the external-agent fallback forever.

### Decision
Issue #600 still requires externally registered agents to be detected and auto-cached on first visit. With this change, that no longer works for users who already have a PlotLink row but registered their ERC-8004 agent elsewhere, so I’m requesting changes.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The lightweight reverse-lookup restores one external-registration case, but it still misses owner-wallet external agents and it does not satisfy issue #600’s zero-RPC requirement for known non-agent users.

### Findings
- [high] Known PlotLink users who externally register as NFT owners with a separate or unset agent wallet are still treated as non-agents, because the known-user path only calls `agentIdByWallet(address)` and never runs the owner lookup fallback.
  - File: `src/app/agents/page.tsx:42`
  - File: `src/app/agents/page.tsx:63`
  - File: `src/components/AgentDashboard.tsx:36`
  - File: `src/components/AgentDashboard.tsx:55`
  - Suggestion: preserve owner-aware fallback for stale known-user rows too, or introduce another way to distinguish “known local user, stale agent cache” from “definitive non-agent”.
- [medium] Known non-agent users still make an RPC call on every `/agents` visit, so the ticket’s “zero RPC calls on /agents page” acceptance criterion is still not met.
  - File: `src/app/agents/page.tsx:42`
  - Suggestion: avoid a per-visit reverse lookup for known non-agents; use an explicit refresh or a cache-invalidation path instead.

### Decision
Issue #600 requires known non-agent users to avoid RPC on `/agents` while still allowing externally registered agents to be detected and auto-cached. This revision still fails both edges, so I’m requesting changes.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The `/agents` fallback logic now matches issue #600 much more closely, but this revision regresses the cached owner dashboard path.

### Findings
- [high] Cached owner dashboards stop loading the agent's storylines when the owner wallet differs from the bound agent wallet, because the DB-backed path no longer swaps `writerAddress` to the cached `agent_wallet`.
  - File: `src/components/AgentDashboard.tsx:80`
  - File: `src/components/AgentDashboard.tsx:107`
  - Suggestion: restore the previous `dbIsOwner && dbAgentWallet` assignment so DB-detected owners query storylines with the cached agent wallet rather than the connected owner wallet.

### Decision
I’m holding approval because this PR introduces a behavioral regression in `AgentDashboard` for owners of agents with separate bound wallets, even though the `/agents` page RPC suppression itself is now on the right track.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The cached owner dashboard regression is fixed, but this PR still removes external-agent detection for existing PlotLink users without actually providing the replacement refresh path it claims.

### Findings
- [high] Existing PlotLink users who later register an ERC-8004 agent elsewhere are now permanently treated as non-agents, because the known-user path suppresses all RPC fallback and the referenced `/api/user/onboard` refresh route does not sync agent cache data.
  - File: `src/app/agents/page.tsx:40`
  - File: `src/components/AgentDashboard.tsx:34`
  - File: `src/app/api/user/onboard/route.ts:71`
  - File: `src/app/api/user/onboard/route.ts:106`
  - Suggestion: either restore a real agent-sync path for stale known users, or make `/api/user/onboard` actually refresh and persist ERC-8004 agent state before treating known users as definitively non-agent.

### Decision
I can’t approve a regression where existing PlotLink users lose external-agent detection entirely. The zero-RPC optimization is fine, but the replacement path cited in the PR is not implemented.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new `/api/user/onboard` sync path is a good direction, but it still only detects direct agent-wallet registrations and misses owner-wallet external agents.

### Findings
- [high] Existing PlotLink users who externally register an agent as the NFT owner with a separate or unset bound wallet still remain invisible after profile refresh, because onboarding only calls `getAgentMetadata(walletAddress)`, which resolves via `agentIdByWallet(wallet)` and returns `null` for owner-only cases.
  - File: `src/app/api/user/onboard/route.ts:118`
  - File: `lib/contracts/erc8004.ts:246`
  - Suggestion: use an owner-aware lookup in the refresh path as well, for example by checking ownership and then resolving metadata by `agentId`.

### Decision
I’m keeping this in request-changes state because the claimed replacement sync path for stale existing users is still incomplete for owner-wallet external registrations.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new onboarding sync path now covers owner-wallet detection, but it still caches direct bound-wallet registrations incorrectly.

### Findings
- [high] Profile refresh caches direct agent-wallet registrations with `agent_wallet = NULL`, because onboarding writes `agentMeta.agentWallet` after calling `getAgentMetadata(wallet)`, and that helper does not populate `agentWallet` for the direct-wallet path. On the next DB-backed `/agents` load, the bound wallet is misclassified as the owner role and gets owner-only management UI.
  - File: `src/app/api/user/onboard/route.ts:121`
  - File: `src/app/api/user/onboard/route.ts:154`
  - File: `lib/contracts/erc8004.ts:271`
  - File: `src/app/agents/page.tsx:77`
  - File: `src/components/AgentManage.tsx:293`
  - Suggestion: when onboarding finds an agent via `getAgentMetadata(normalizedAddress)`, persist `agent_wallet: normalizedAddress` directly, or have `getAgentMetadata()` return `agentWallet: walletAddress` for the bound-wallet case.

### Decision
I’m keeping this in request-changes state because the explicit refresh path still corrupts role detection for existing PlotLink users who externally register as direct agent wallets.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The `/agents` page now skips RPC for known non-agent users while preserving full fallback for unknown wallets, and the new onboarding sync path covers both bound-wallet and owner-wallet external registrations without breaking DB-backed role detection.

### Findings
- None.

### Decision
This now satisfies issue #600’s acceptance criteria: known non-agents stay zero-RPC on page load, known agents remain DB-backed, unknown wallets still use RPC fallback, and external registrations have a concrete sync path that correctly populates the agent cache.

PR #618, opened 2026-03-27
Fix mobile input auto-zoom on iOS Safari

### LINKED ISSUE #612: Fix mobile input auto-zoom — add maximumScale: 1 to viewport
## Bug

On mobile (iOS Safari), tapping any input field triggers an auto-zoom that doesn't zoom back out. This happens because iOS zooms in on inputs with `font-size < 16px`, and the browser doesn't restore the original zoom level after the input loses focus.

## Root Cause

PlotLink's viewport config in `src/app/layout.tsx` (line 32-36) is missing `maximumScale: 1`:

```typescript
export const viewport: Viewport = {
  width: "device-width",
  initialScale: 1,
  viewportFit: "cover",
  // missing: maximumScale: 1
};
```

## Fix

Same fix as Dropcast (`~/Projects/dropcast/app/layout.tsx` line 62-65) — add `maximumScale: 1`:

```typescript
export const viewport: Viewport = {
  width: "device-width",
  initialScale: 1,
  maximumScale: 1,
  viewportFit: "cover",
};
```

This prevents iOS Safari from auto-zooming on input focus.

## Files to modify

- `src/app/layout.tsx` — add `maximumScale: 1` to viewport export

## Branch

`task/612-mobile-input-zoom`

## Acceptance criteria

- [ ] Input fields no longer trigger auto-zoom on iOS Safari
- [ ] Viewport stays at 1x scale after interacting with inputs
- [ ] Build passes

realproject7:
## Summary
Fixes #612

- Add `maximumScale: 1` to the Next.js `Viewport` export in `layout.tsx`
- Prevents iOS Safari from auto-zooming when focusing input/textarea elements

## Test plan
- [ ] iOS Safari: tapping an input field does not zoom the page
- [ ] Desktop: no layout change
- [ ] Build passes

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -203,6 +203,13 @@ code, pre, .font-mono {
   }
 }
 
+/* Prevent iOS Safari auto-zoom on input focus (triggers at font-size < 16px) */
+@supports (-webkit-touch-callout: none) {
+  input, textarea, select {
+    font-size: max(16px, 1em);
+  }
+}
+
 /* Custom select dropdown styling */
 select {
   appearance: none;
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -32,6 +32,7 @@ const themeColor = "#E8DFD0";
 export const viewport: Viewport = {
   width: "device-width",
   initialScale: 1,
+  maximumScale: 1,
   viewportFit: "cover",
 };
 
--- a/src/app/layout.tsx
+++ b/src/app/layout.tsx
@@ -32,7 +32,6 @@ const themeColor = "#E8DFD0";
 export const viewport: Viewport = {
   width: "device-width",
   initialScale: 1,
-  maximumScale: 1,
   viewportFit: "cover",
 };
 


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The change is scoped correctly to `layout.tsx`, but the chosen fix applies a global zoom cap to every mobile browser instead of addressing only the iOS Safari auto-zoom behavior.

### Findings
- [high] `maximumScale: 1` disables user zoom on browsers that honor the viewport limit, which turns an iOS input-focus annoyance into an accessibility regression for the whole app.
  - File: `src/app/layout.tsx:35`
  - Suggestion: avoid a global viewport zoom cap and use a targeted fix for the affected inputs instead, such as ensuring mobile form controls render at `16px`+ or another iOS-specific approach that does not remove pinch zoom.

### Decision
Requesting changes because this global viewport restriction is too broad for the bug being fixed and can break zoom accessibility on mobile browsers outside the original iOS Safari case.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR now uses a targeted CSS workaround for the iOS Safari input auto-zoom trigger instead of a global viewport zoom cap.

### Findings
- None.

### Decision
Approved because the updated fix addresses issue #612 without removing pinch-zoom for the whole app: `src/app/globals.css` now raises form-control font sizes to at least 16px in the targeted Safari path, which is the actual condition that triggers the unwanted auto-zoom.

PR #309, opened 2026-03-18
[#308] GitHub-style inline filter bar with clickable params

realproject7:
## Summary
- New `FilterBar` client component replacing `WriterFilter`, `GenreFilter`, `LanguageFilter`, and `SortDropdown`
- Single bordered box with inline `label:value` tokens — values in green accent
- Clicking a token opens a positioned dropdown below it; only one open at a time
- Outside click closes dropdown
- Mobile: sort collapsed to `↕` icon to fit everything on one line
- Desktop: sort shows full `sort:Recent` label
- Same URL query param behavior (`tab`, `writer`, `genre`, `lang`) preserved
- Removed unused `extraParams` variable

Fixes #308

## Test plan
- [ ] Verify all filters appear as inline tokens in one bordered box
- [ ] Verify values are green accent color
- [ ] Click each token — dropdown opens below with correct options
- [ ] Only one dropdown open at a time
- [ ] Click outside closes dropdown
- [ ] Mobile: sort shows icon only, all tokens fit one line
- [ ] Selecting a filter navigates with correct query params
- [ ] Desktop layout looks clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #646, opened 2026-03-30
[#638] Fix PLOT-mode MAX buy path and remove unnecessary delays

### LINKED ISSUE #638: Fix MAX button broken in PLOT mode + remove unnecessary 5s delays
### 1. MAX button broken in PLOT mode

In \`TradingWidget.tsx\`, \`handleBuyMax\` always calls \`getZapQuote()\` regardless of whether the user is buying with PLOT directly. When \`payToken === "PLOT"\` and Zap is available, \`isPlotMode\` is false so it tries the Zap quote — which silently fails since PLOT is the intermediate token, not a Zap input. The MAX button does nothing in PLOT mode.

**Fix:** Check \`isPlotMode\` in \`handleBuyMax\` and use the direct bonding curve quote instead of Zap when buying with PLOT.

### 2. Unnecessary 5-second delays after tx confirmation

Both \`DonateWidget.tsx\` (line 94) and \`usePublish.ts\` (line 116) have:
\`\`\`typescript
await new Promise((r) => setTimeout(r, 5000));
\`\`\`

This was for "RPC propagation on Base Sepolia" but is unnecessary on mainnet Base with 2-second block times. The receipt is already confirmed before this delay runs.

**Fix:** Remove the 5-second delays. If any propagation buffer is needed, 1 second is sufficient.

## Files to modify

- \`src/components/TradingWidget.tsx\` — fix \`handleBuyMax\` for PLOT mode
- \`src/components/DonateWidget.tsx\` — remove/reduce 5s delay
- \`src/hooks/usePublish.ts\` — remove/reduce 5s delay

## Branch

\`task/638-ux-fixes\`

## Acceptance criteria

- [ ] MAX button works correctly when buying with PLOT
- [ ] No 5-second delays after tx confirmation
- [ ] Build passes

---

## Self-Verification (T3)

- [ ] Run `npm run dev`, go to a story page, select PLOT as pay token
- [ ] Click MAX button — verify it fills in the correct max amount (not silent failure)
- [ ] Create a storyline — verify no 5-second delay after tx confirmation before indexing
- [ ] Make a donation — verify no 5-second delay after tx confirmation

realproject7:
## Summary
- **TradingWidget.tsx**: `handleBuyMax` no longer calls `getZapQuote()` in PLOT mode (which silently failed). Instead, uses binary search over `getReserveForToken` on the bonding curve to find the maximum storyline tokens mintable within the user's PLOT balance.
- **DonateWidget.tsx**: Removed 5-second `setTimeout` delay before indexer call — unnecessary on Base mainnet.
- **usePublish.ts**: Removed 5-second `setTimeout` delay before indexer call — unnecessary on Base mainnet.

Fixes realproject7/plotlink#638
Tracks realproject7/agent-os#309

## Test plan
- [ ] In PLOT mode, clicking MAX successfully prepares the buy using the bonding-curve quote path
- [ ] In non-PLOT mode (ETH/USDC/HUNT), clicking MAX still uses the zap flow as before
- [ ] Donate flow proceeds without the extra 5-second delay
- [ ] Publish flow proceeds without the extra 5-second delay

### DIFF
--- a/src/components/DonateWidget.tsx
+++ b/src/components/DonateWidget.tsx
@@ -89,9 +89,7 @@ export function DonateWidget({ storylineId, writerAddress }: DonateWidgetProps) 
       setTxState("pending");
       await publicClient.waitForTransactionReceipt({ hash });
 
-      // Trigger donation indexer (delay for RPC propagation on Base Sepolia)
       setTxState("indexing");
-      await new Promise((r) => setTimeout(r, 5000));
       const indexRes = await fetch("/api/index/donation", {
         method: "POST",
         headers: { "Content-Type": "application/json" },
--- a/src/components/TradingWidget.tsx
+++ b/src/components/TradingWidget.tsx
@@ -180,34 +180,92 @@ export function TradingWidget({ tokenAddress }: { tokenAddress: Address }) {
       if (maxBalance <= BigInt(0)) return;
 
       if (isPlotMode) {
-        // PLOT mode: binary search for max storyline tokens mintable within PLOT balance
-        let lo = BigInt(1);
-        let hi = maxBalance; // upper bound: 1 storyline token per 1 PLOT (always overshoots)
-        let best = BigInt(0);
+        // PLOT mode: find max storyline tokens mintable within PLOT balance.
+        // Uses batched multicall probes to minimize RPC round-trips (2-4 calls total).
 
-        for (let i = 0; i < 64; i++) {
-          if (lo > hi) break;
-          const mid = (lo + hi) / BigInt(2);
-          try {
-            const [reserveNeeded] = await publicClient.readContract({
-              address: MCV2_BOND,
-              abi: mcv2BondAbi,
-              functionName: "getReserveForToken",
-              args: [tokenAddress, mid],
-            }) as [bigint, bigint];
+        // Step 1: Exponential search to find upper bound (tokens can be very cheap early on the curve)
+        const expProbes: bigint[] = [];
+        for (let exp = BigInt(0); exp < BigInt(20); exp++) {
+          expProbes.push(BigInt(10) ** exp * BigInt(10) ** BigInt(18)); // 1, 10, 100, ... × 1e18
+        }
+
+        const expResults = await publicClient.multicall({
+          contracts: expProbes.map((probe) => ({
+            address: MCV2_BOND,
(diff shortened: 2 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR fixes the silent PLOT-mode failure and removes the artificial waits, but the new PLOT MAX path still does not reliably compute the actual maximum buyable amount.

### Findings
- [high] The binary-search upper bound assumes `1 storyline token <= 1 PLOT`, which is not guaranteed by the bonding curve.
  - File: `src/components/TradingWidget.tsx:185`
  - Suggestion: Grow `hi` until `getReserveForToken(tokenAddress, hi)` exceeds the user's PLOT balance (for example by doubling from 1 token), then binary-search within that bracket. That keeps MAX correct when the marginal price is below 1 PLOT and the user can mint more than `maxBalance` token units.

### Decision
Requesting changes because the current implementation can still underfill MAX buys in PLOT mode, so it does not meet the acceptance criterion that MAX works correctly.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The revised multicall search fixes the earlier upper-bound / round-trip issues, but the PLOT-mode MAX path still misses a valid buy range below one full token.

### Findings
- [high] The probe ladder starts at `1e18` and returns early when that first whole-token probe is too expensive, so MAX still does nothing for wallets that can afford only a fractional token.
  - File: `src/components/TradingWidget.tsx:188`
  - Suggestion: Include sub-1-token probes (for example `1e17`, `1e16`, ...), or fall back to a finer-grained search in `(0, 1e18)` when the first probe exceeds `maxBalance`. The widget already accepts 18-decimal token amounts, so MAX needs to cover that range too.

### Decision
Requesting changes because users who can buy less than one storyline token still get a broken MAX button, which does not satisfy the acceptance criterion that MAX works correctly in PLOT mode.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The requested fixes are addressed. PLOT-mode MAX now uses the bonding-curve path instead of Zap, covers fractional token amounts, and the 5-second waits remain removed from donate and publish.

### Findings
- None.

### Decision
Approving on code review. `lint-and-typecheck` is passing; the remaining checks are still running, but they do not block the review verdict.

Issue #312, opened 2026-03-18
[chore] Clean up all ESLint errors and warnings


realproject7:
## Summary

Clean up all 8 pre-existing ESLint issues so `npm run lint` passes clean.

### 1. Exclude SDK dist from linting (4 errors)

`packages/sdk/dist/index.cjs` triggers `@typescript-eslint/no-require-imports` — this is compiled output and should not be linted.

**Fix**: Add `packages/sdk/dist/` to ESLint ignore config (`.eslintignore` or `ignores` in `eslint.config.*`).

### 2. Remove unused `reset` destructuring (1 warning)

`src/app/create/page.tsx:50` — `reset` is destructured from `usePublish()` but never used.

**Fix**: Remove `reset` from the destructuring.

### 3. Remove unnecessary `tokenAddress` from useCallback deps (1 warning)

`src/components/ClaimRoyalties.tsx:87` — `tokenAddress` is a prop (stable reference from parent), not a reactive value. ESLint correctly flags it as unnecessary.

**Fix**: Remove `tokenAddress` from the dependency array.

### 4. Wrap `allOptions` in useMemo (1 warning)

`src/components/Select.tsx:32` — `allOptions` is recomputed every render, causing the `handleKeyDown` useCallback deps to change every render, defeating memoization.

**Fix**: Wrap `allOptions` in `useMemo(() => ..., [placeholder, options])`.

### 5. Unused expression in SDK dist (1 warning)

`packages/sdk/dist/index.cjs:603` — same file as #1, will be resolved by excluding dist from linting.

## Acceptance Criteria

- [ ] `npm run lint` exits with 0 errors and 0 warnings
- [ ] `npm run typecheck` still passes
- [ ] No functional changes to app behavior

## Labels
`agent/T3`, `plotlink`, `chore`

PR #649, opened 2026-03-30
[#639] Harden resolveAgentURI with timeout and validation

### LINKED ISSUE #639: Add timeout and validation for remote agent URI fetches
## Problem

In \`lib/contracts/erc8004.ts\`, \`resolveAgentURI()\` fetches remote URLs without timeout, size limit, or response validation:

\`\`\`typescript
const res = await fetch(fetchUrl);
return (await res.json()) as Record<string, unknown>;
\`\`\`

A malicious agent could register a URI pointing to a very large file, slow-responding server, or non-JSON content — causing resource exhaustion or hangs on the PlotLink server.

## Fix

1. Add a fetch timeout (e.g., 5 seconds via \`AbortController\`)
2. Check \`res.ok\` before parsing
3. Limit response size (e.g., read first 50KB only)
4. Wrap in try/catch for JSON parse errors
5. Also add size check for \`data:\` URI payloads

## Files to modify

- \`lib/contracts/erc8004.ts\` — \`resolveAgentURI()\` function

## Branch

\`task/639-agent-uri-validation\`

## Acceptance criteria

- [ ] Remote fetches have 5-second timeout
- [ ] Response status checked before parsing
- [ ] Response size limited
- [ ] JSON parse errors handled gracefully
- [ ] data: URI payloads size-checked
- [ ] Build passes

---

## Self-Verification (T3)

- [ ] Write a unit test or manual test: call `resolveAgentURI('https://httpstat.us/200?sleep=10000')` — verify it times out after 5 seconds
- [ ] Call `resolveAgentURI('https://httpstat.us/500')` — verify it returns null (not crash)
- [ ] Call `resolveAgentURI('not-json-content')` — verify graceful error handling
- [ ] Call `resolveAgentURI('data:text/plain,...<50KB+>')` — verify size check
- [ ] Normal agent URI still resolves correctly

realproject7:
## Summary
- **5s timeout**: `AbortController` on remote fetches (https/ipfs) prevents hangs from slow servers
- **Response validation**: Checks `res.ok` before parsing body
- **50KB size limit**: Enforced on all payload types (remote, `data:` URI, raw JSON)
- **Graceful error handling**: Entire function wrapped in try/catch, returns `{}` on any failure instead of throwing

Fixes realproject7/plotlink#639
Tracks realproject7/agent-os#312

## Test plan
- [ ] Remote fetch with slow server times out after 5s (returns `{}`)
- [ ] Non-200 responses return `{}` (not crash)
- [ ] Invalid JSON returns `{}` gracefully
- [ ] Oversized `data:` URI payloads return `{}`
- [ ] Normal agent URIs still resolve correctly
- [ ] Build passes

### DIFF
--- a/lib/contracts/erc8004.ts
+++ b/lib/contracts/erc8004.ts
@@ -212,27 +212,48 @@ export async function detectWriterType(
   }
 }
 
+const MAX_URI_BYTES = 50 * 1024; // 50 KB size limit
+const FETCH_TIMEOUT_MS = 5000; // 5-second timeout
+
 /**
  * Resolve an agent URI string to a parsed JSON object.
  * Handles raw JSON, data: URIs (base64 + URL-encoded), https://, and ipfs://.
+ * Includes timeout, size limits, and response validation for safety.
  */
 export async function resolveAgentURI(uri: string): Promise<Record<string, unknown>> {
-  if (uri.startsWith("{")) {
-    return JSON.parse(uri);
+  try {
+    if (uri.startsWith("{")) {
+      if (uri.length > MAX_URI_BYTES) return {};
+      return JSON.parse(uri);
+    }
+    if (uri.startsWith("data:")) {
+      const comma = uri.indexOf(",");
+      const payload = comma >= 0 ? uri.slice(comma + 1) : uri;
+      if (payload.length > MAX_URI_BYTES) return {};
+      return JSON.parse(
+        uri.includes("base64") ? atob(payload) : decodeURIComponent(payload),
+      );
+    }
+    // https:// or ipfs://
+    const fetchUrl = uri.startsWith("ipfs://")
+      ? uri.replace("ipfs://", "https://ipfs.io/ipfs/")
+      : uri;
+
+    const controller = new AbortController();
+    const timeout = setTimeout(() => controller.abort(), FETCH_TIMEOUT_MS);
+    try {
+      const res = await fetch(fetchUrl, { signal: controller.signal });
+      if (!res.ok) return {};
+
+      const text = await res.text();
+      if (text.length > MAX_URI_BYTES) return {};
+      return JSON.parse(text);
+    } finally {
+      clearTimeout(timeout);
+    }
+  } catch {
+    return {};
   }
-  if (uri.startsWith("data:")) {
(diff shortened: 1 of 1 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The timeout and graceful-failure work is headed in the right direction, but the oversized remote-response protection is still incomplete.

### Findings
- [high] The remote size check happens only after `res.text()` has already buffered the entire response body, so an oversized agent URI response can still be fully downloaded into memory before being rejected.
  - File: `lib/contracts/erc8004.ts:248`
  - Suggestion: Reject oversized remote payloads before or during body consumption. At minimum, gate on `Content-Length` when present; ideally stream the body and abort once the accumulated byte count exceeds `MAX_URI_BYTES`.

### Decision
Requesting changes because the current implementation does not yet safely handle oversized remote responses, which is part of the issue's acceptance criteria.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The streamed size-cap fix addresses the earlier buffering issue, but the current implementation introduces a client-side runtime/bundle regression.

### Findings
- [high] `resolveAgentURI()` now uses `Buffer.concat(...)`, but this helper is imported by the client component `src/components/AgentManage.tsx`, so the change adds a Node-only dependency into client code.
  - File: `lib/contracts/erc8004.ts:267`
  - Suggestion: Keep the streamed decoding browser-safe. For example, accumulate into a growing `Uint8Array`, or append decoded chunks via `TextDecoder.decode(value, { stream: true })` and finalize once the reader completes.

### Decision
Requesting changes because `resolveAgentURI()` is used from client-side code and the new `Buffer` dependency can break the runtime or bundle despite the size-limit fix itself being directionally correct.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The hardening work now meets the requested scope: `resolveAgentURI()` adds the timeout, response validation, size limits, graceful parse failure handling, and the streamed body handling is now browser-safe for client-importable code.

### Findings
- None.

### Decision
Approving on code review. The earlier oversized-response and client-runtime issues are addressed; remaining CI checks are still running but do not block the review verdict.

PR #313, opened 2026-03-18
[#312] Clean up all ESLint errors and warnings

realproject7:
## Summary
- Excluded `packages/sdk/dist/` from ESLint config (resolves 4 errors + 1 warning from compiled output)
- Removed unused `reset` destructuring in `src/app/create/page.tsx`
- Removed unnecessary `tokenAddress` dependency in `src/components/ClaimRoyalties.tsx`
- Wrapped `allOptions` in `useMemo` in `src/components/Select.tsx`
- `npm run lint` now exits with 0 errors and 0 warnings

Fixes #312

## Test plan
- [ ] `npm run lint` exits cleanly (0 errors, 0 warnings)
- [ ] `npm run typecheck` passes
- [ ] `npm run build` passes
- [ ] No functional behavior changes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #650, opened 2026-03-30
[#633] Unify user upsert to fix concurrent race condition

### LINKED ISSUE #633: Fix register-by-wallet vs onboard upsert race condition
## Problem

`register-by-wallet` and `onboard` are two separate API routes that both upsert the same `users` row. When called concurrently (e.g., wallet connect triggers register while user clicks refresh), they race:

1. Both query for existing user separately
2. Both attempt INSERT — first succeeds, second gets 23505 (unique violation)
3. Second falls back to UPDATE but uses different conflict keys (FID vs primary_address)
4. Result: partial updates, corrupted row with mixed data from both routes

## Root Cause

Both routes use INSERT-then-catch-23505-then-UPDATE pattern with different conflict resolution:
- `register-by-wallet`: updates by `fid` or `primary_address`
- `onboard`: updates by `existingUser.id`

## Fix

1. **Unify the upsert logic** — both routes should use the same conflict resolution strategy, preferably `.upsert()` with explicit conflict keys
2. **Use `existingUser.id`** as the primary update key in both routes (most reliable — UUID, always unique)
3. **Add a shared `upsertUser()` helper** in `lib/actions.ts` that both routes call, ensuring consistent behavior
4. Handle all user types correctly: FID users, wallet-only users, agent users

## Files to modify

- `src/app/api/user/register-by-wallet/route.ts` — use shared upsert logic
- `src/app/api/user/onboard/route.ts` — use shared upsert logic
- `lib/actions.ts` or new `lib/user-upsert.ts` — shared upsert helper

## Branch

`task/633-upsert-race-fix`

## Acceptance criteria

- [ ] Concurrent register + onboard calls don't corrupt user row
- [ ] Both routes use identical conflict resolution
- [ ] FID users, wallet-only users, and agent users all handled correctly
- [ ] Build passes

---

## Self-Verification (T3)

- [ ] Run `npm run dev`, connect a NEW wallet (not in DB yet)
- [ ] Simultaneously trigger register-by-wallet + onboard (connect wallet then immediately click Refresh Profile)
- [ ] Check Supabase — verify user row is consistent (not duplicated, no mixed data)
- [ ] Repeat 3 times — no 500 errors in server logs
- [ ] Run `npm run typecheck` — no type errors

realproject7:
## Summary
- **New `lib/user-upsert.ts`**: Shared `findUserByWallet()` and `upsertUser()` helpers with consistent conflict resolution — always updates by row `id`, re-queries on 23505 conflict to find the correct row.
- **register-by-wallet**: Replaced inline lookup + INSERT-then-catch-UPDATE with shared helpers.
- **onboard**: Same replacement — both routes now use identical upsert logic.

Previously, concurrent calls used different conflict keys (`fid` vs `primary_address`) which could corrupt rows with mixed data.

Fixes realproject7/plotlink#633
Tracks realproject7/agent-os#313

## Test plan
- [ ] Concurrent register + onboard calls don't corrupt or duplicate user rows
- [ ] FID users, wallet-only users, and agent users all upsert correctly
- [ ] 5-min cooldown still enforced on onboard route
- [ ] Build passes

### DIFF
--- a/lib/user-upsert.ts
+++ b/lib/user-upsert.ts
@@ -10,7 +10,9 @@ type UserInsert = Database["public"]["Tables"]["users"]["Insert"];
 type UserRow = Database["public"]["Tables"]["users"]["Row"];
 
 /**
- * Find an existing user by wallet address (checks verified_addresses then primary_address).
+ * Find an existing user by wallet address.
+ * Checks verified_addresses, primary_address, agent_wallet, and agent_owner
+ * to match the full lookup chain used by getUserFromDB().
  */
 export async function findUserByWallet(
   supabase: SupabaseClient<Database>,
@@ -30,7 +32,23 @@ export async function findUserByWallet(
     .eq("primary_address", normalizedAddress)
     .single();
 
-  return byPrimary;
+  if (byPrimary) return byPrimary;
+
+  const { data: byAgentWallet } = await supabase
+    .from("users")
+    .select("*")
+    .eq("agent_wallet", normalizedAddress)
+    .single();
+
+  if (byAgentWallet) return byAgentWallet;
+
+  const { data: byAgentOwner } = await supabase
+    .from("users")
+    .select("*")
+    .eq("agent_owner", normalizedAddress)
+    .single();
+
+  return byAgentOwner ?? null;
 }
 
 /**
--- a/lib/user-upsert.ts
+++ b/lib/user-upsert.ts
@@ -1,0 +1,88 @@
+/**
+ * Shared user upsert logic for register-by-wallet and onboard routes.
+ * Ensures consistent conflict resolution so concurrent calls cannot corrupt rows.
+ */
+
+import type { SupabaseClient } from "@supabase/supabase-js";
+import type { Database } from "./supabase";
+
+type UserInsert = Database["public"]["Tables"]["users"]["Insert"];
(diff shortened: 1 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The shared upsert refactor is directionally right, but the new lookup helper regresses agent-user handling.

### Findings
- [high] `findUserByWallet()` only checks `verified_addresses` and `primary_address`, so both routes now miss existing rows that are keyed by `agent_wallet` / `agent_owner`.
  - File: `lib/user-upsert.ts:16`
  - Suggestion: Reuse the full wallet lookup behavior from `getUserFromDB()` (including `agent_wallet` and `agent_owner`), or otherwise extend `findUserByWallet()` so agent users participate in the same cooldown and id-based upsert path.

### Decision
Requesting changes because the issue explicitly requires preserving correct handling for agent users, and the new helper can return `null` for those rows, which breaks the conflict-retry/update path in both routes.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The shared upsert refactor now preserves the broader wallet lookup behavior for agent users while unifying both routes onto the same id-based conflict resolution path.

### Findings
- None.

### Decision
Approving on code review. The earlier agent-user regression is addressed; the remaining CI checks are still running, but they do not block the review verdict.

PR #654, opened 2026-03-30
[#637] Align trade price calculation between real-time and cron indexers

### LINKED ISSUE #637: Fix price inconsistency between real-time and cron trade indexers
## Problem

The same trade gets different \`price_per_token\` values depending on which indexer path processes it:

- **Real-time indexer** (\`/api/index/trade\`): Uses \`priceForNextMint\` at the historical block for marginal bonding curve price, falling back to batch average
- **Cron indexer** (\`/api/cron/trade-history\`): Always uses batch average (\`reserveAmount / tokenAmount\`)

This creates inconsistent price charts and trading history depending on timing.

## Fix

Both indexers should use the same price calculation method. Recommend using batch average (\`reserveAmount / tokenAmount\`) consistently since:
1. It reflects actual cost paid (not marginal price)
2. It doesn't require an extra RPC call
3. It's deterministic

Or use marginal price in both — but be consistent.

## Files to modify

- \`src/app/api/index/trade/route.ts\` — align price calculation
- \`src/app/api/cron/trade-history/route.ts\` — align price calculation

## Branch

\`task/637-price-consistency\`

## Acceptance criteria

- [ ] Both indexers use the same price calculation method
- [ ] Existing price data is consistent going forward
- [ ] Build passes

---

## Self-Verification (T3)

- [ ] Read both indexer files and confirm they use the same price calculation method
- [ ] Run `npm run dev`, make a trade — check `trade_history` table for `price_per_token` value
- [ ] Trigger the cron trade-history for the same tx — verify `price_per_token` matches

realproject7:
## Summary
- **Real-time indexer** (`/api/index/trade`): Replaced `priceForNextMint` (marginal bonding curve price) with batch average (`reserveAmount / tokenAmount`) to match the cron indexer.
- Removed unused `priceForNextMintFunction` import and extra RPC call.
- Both indexers now produce identical `price_per_token` values for the same trade.

Fixes realproject7/plotlink#637
Tracks realproject7/agent-os#317

## Test plan
- [ ] Both indexers use same price calculation (batch average)
- [ ] Trade history `price_per_token` is consistent regardless of which indexer runs
- [ ] No extra `priceForNextMint` RPC call in real-time path
- [ ] Build passes

### DIFF
--- a/src/app/api/index/trade/route.ts
+++ b/src/app/api/index/trade/route.ts
@@ -2,7 +2,7 @@ import { NextResponse } from "next/server";
 import { type Hex, decodeEventLog, formatUnits } from "viem";
 import { publicClient, getReceiptWithRetry } from "../../../../../lib/rpc";
 import { createServerClient } from "../../../../../lib/supabase";
-import { mcv2BondEventAbi, priceForNextMintFunction } from "../../../../../lib/contracts/abi";
+import { mcv2BondEventAbi } from "../../../../../lib/contracts/abi";
 import { MCV2_BOND, ZAP_PLOTLINK } from "../../../../../lib/contracts/constants";
 import { erc20Abi } from "../../../../../lib/price";
 import { validateRecentTx } from "../../../../../lib/index-auth";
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
## Verdict: APPROVE

### Summary
The realtime trade indexer now uses the same batch-average `price_per_token` calculation as the cron trade-history route, so the two paths will stop producing divergent values for the same trade. The change is minimal and stays within the scope of the ticket.

### Findings
- No blocking findings.

### Decision
Approving because the updated realtime calculation matches the cron path and resolves the consistency issue described in Queue #637 / agent-os#317.

Issue #318, opened 2026-03-18
[ui] Add 'Read the first Plot' strip button on story detail page


realproject7:
## Summary

Add a full-width "Read the first Plot" button below the story header on the detail page. It should scroll/anchor down to the genesis plot section.

**File**: `src/app/story/[storylineId]/page.tsx`

## Acceptance Criteria
- [ ] Full-width green accent outlined button below the header
- [ ] Clicking scrolls to the genesis section
- [ ] Only shown when a genesis plot exists

## Labels
`agent/T3`, `plotlink`, `ui`

PR #655, opened 2026-03-30
[#630] Two-phase book page turn with visible page underneath

### LINKED ISSUE #630: Reading Mode — realistic book page turn effect with visible page underneath
## Follow-up from PR #629

The current page flip uses a simple `rotateY(-90deg)` which rotates the content off-screen. We want a more realistic **book page turning** effect where:

1. The outgoing page visually lifts and curls from the edge like a physical paper page
2. The incoming (next) page is visible **underneath** as the outgoing page turns
3. A shadow follows the turning page to add depth

## Reference

See the attached screenshot of a PDF reader — the page physically turns and reveals the next page behind it, with shadows along the fold line.

### Two-phase page turn:

1. **Phase 1 (outgoing page flips away):** The current page starts flat, then rotates around the Y-axis from the right edge (for Next) or left edge (for Prev). As it rotates, a shadow appears along the fold. The page should have `backface-visibility: hidden` so you don't see reversed content.

2. **Phase 2 (incoming page revealed):** The next page content is already positioned underneath (z-index behind). As the outgoing page lifts away, the incoming page is progressively revealed. The incoming page can have a subtle fade-in or slide-in from slightly offset.

### Implementation approach:

Stack both pages (outgoing on top, incoming below):
```css
.page-container {
  position: relative;
  perspective: 1500px;
}

.page-outgoing {
  position: absolute;
  transform-origin: left center;  /* for "Next" */
  animation: flip-out 500ms ease-in forwards;
  backface-visibility: hidden;
  z-index: 2;
}

.page-incoming {
  z-index: 1;
  animation: fade-in 500ms ease-out;
}

@keyframes flip-out {
  0% { transform: rotateY(0deg); box-shadow: none; }
  50% { box-shadow: -10px 0 30px rgba(0,0,0,0.2); }
  100% { transform: rotateY(-120deg); box-shadow: none; }
}
```

### Key details:
- Both pages must be in the DOM simultaneously during the transition
- The outgoing page captures the previous content (can use a snapshot or keep the previous content in state)
- The shadow should move with the page fold
- Duration: ~500ms ease-in-out
- Must work on both swipe and button/keyboard navigation
- CSS-only animation (no JS animation libraries)
- Mobile performance: use `will-change: transform` and GPU-accelerated properties only

## Files to modify

- `src/components/ReadingMode.tsx` — two-phase render with outgoing/incoming pages
- `src/app/globals.css` — updated keyframes with shadow and backface

## Branch

`task/623-reading-mode-page-flip-v2` (same branch, follow-up)

## Acceptance criteria

- [ ] Outgoing page visually lifts and turns like a real book page
- [ ] Incoming page visible underneath during the turn
- [ ] Shadow along the fold line during transition
- [ ] Backface of turning page is hidden (no reversed text)
- [ ] Works on swipe, button click, and keyboard arrows
- [ ] Smooth on mobile (GPU-accelerated transforms only)
- [ ] No external dependencies
- [ ] Build passes

---

## Self-Verification (T3)

- [ ] Run `npm run dev`, navigate to a story with multiple plots
- [ ] Click Next — verify outgoing page flips and incoming page is visible underneath
- [ ] Click Prev — verify reverse flip direction
- [ ] Swipe left/right — verify same flip effect
- [ ] Press arrow keys — verify same flip effect
- [ ] Check mobile (responsive mode) — verify smooth performance
- [ ] Verify `backface-visibility: hidden` — no reversed text visible during flip

realproject7:
## Summary
- **ReadingMode.tsx**: Replaced sequential flip-out/in with simultaneous two-page stack. Both outgoing and incoming pages are in the DOM during the transition. Outgoing page (previous content) is absolutely positioned on top and flips away, revealing the incoming page underneath.
- **globals.css**: Updated animations — 500ms flip with fold shadow, 120deg rotation for realistic curl, `backface-visibility: hidden`, and incoming page fade-in.

Fixes realproject7/plotlink#630
Tracks realproject7/agent-os#318

## Test plan
- [ ] Click Next — outgoing page flips and incoming page visible underneath
- [ ] Click Prev — reverse flip direction
- [ ] Swipe left/right — same flip effect
- [ ] Arrow keys — same flip effect
- [ ] No reversed text visible during flip (backface hidden)
- [ ] Shadow along fold line during transition
- [ ] Smooth on mobile (GPU-accelerated transforms only)
- [ ] Build passes

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -210,45 +210,63 @@ code, pre, .font-mono {
   }
 }
 
-/* Reading Mode page-flip transitions (3D book-style) */
+/* Reading Mode — two-phase book page turn */
 .page-flip-container {
-  perspective: 1200px;
+  perspective: 1500px;
 }
 
+.page-flip-stack {
+  position: relative;
+  min-height: 100%;
+}
+
+.page-flip-page {
+  position: relative;
+  will-change: transform, opacity;
+}
+
+/* Outgoing page stacks on top and flips away */
+.page-outgoing {
+  position: absolute;
+  inset: 0;
+  z-index: 2;
+  backface-visibility: hidden;
+  overflow: hidden;
+}
+
+/* Incoming page: subtle fade-in from underneath */
+.page-incoming {
+  animation: page-reveal 500ms ease-out;
+}
+
+@keyframes page-reveal {
+  0%   { opacity: 0.4; }
+  60%  { opacity: 0.8; }
+  100% { opacity: 1; }
+}
+
+/* Next: outgoing page flips from right edge (like turning a book page forward) */
 @keyframes flip-out-forward {
-  0%   { transform: rotateY(0deg); opacity: 1; }
-  100% { transform: rotateY(-90deg); opacity: 0; }
-}
-@keyframes flip-out-backward {
-  0%   { transform: rotateY(0deg); opacity: 1; }
-  100% { transform: rotateY(90deg); opacity: 0; }
-}
-@keyframes flip-in-from-right {
-  0%   { transform: rotateY(90deg); opacity: 0; }
-  100% { transform: rotateY(0deg); opacity: 1; }
-}
-@keyframes flip-in-from-left {
-  0%   { transform: rotateY(-90deg); opacity: 0; }
-  100% { transform: rotateY(0deg); opacity: 1; }
+  0%   { transform: rotateY(0deg); box-shadow: -4px 0 16px rgba(0,0,0,0.1); }
+  40%  { box-shadow: -10px 0 30px rgba(0,0,0,0.2); }
+  100% { transform: rotateY(-120deg); box-shadow: none; }
 }
 
-/* Outgoing page: flips away */
+/* Prev: outgoing page flips from left edge (like turning a book page backward) */
+@keyframes flip-out-backward {
+  0%   { transform: rotateY(0deg); box-shadow: 4px 0 16px rgba(0,0,0,0.1); }
+  40%  { box-shadow: 10px 0 30px rgba(0,0,0,0.2); }
+  100% { transform: rotateY(120deg); box-shadow: none; }
+}
+
 .page-flip-out-left {
   transform-origin: left center;
-  animation: flip-out-forward 200ms ease-in forwards;
+  animation: flip-out-forward 500ms ease-in forwards;
 }
+
 .page-flip-out-right {
   transform-origin: right center;
-  animation: flip-out-backward 200ms ease-in forwards;
-}
-/* Incoming page: flips into view */
-.page-flip-in-left {
-  transform-origin: left center;
-  animation: flip-in-from-right 200ms ease-out;
-}
-.page-flip-in-right {
-  transform-origin: right center;
-  animation: flip-in-from-left 200ms ease-out;
+  animation: flip-out-backward 500ms ease-in forwards;
 }
 
 /* Custom select dropdown styling */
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new two-page stack is the right direction, but the navigation timing now resets the scroll container before the outgoing page animates. That creates a visible jump whenever the reader is not already at the top of the chapter.

### Findings
- [high] `scrollToTop()` now runs immediately after `setCurrentIdx(idx)`, while the outgoing page is still being rendered from the same scrollable container. If the reader is partway through a chapter, the old page will snap back to the top before it flips away, which breaks the page-turn effect across button, swipe, and keyboard navigation.
  - File: `src/components/ReadingMode.tsx:55`
  - Suggestion: Delay the scroll reset until after the outgoing page animation finishes, or otherwise decouple the outgoing page snapshot from the live scroll position before starting the turn.

### Decision
Requesting changes because this revision regresses the core reading interaction for normal mid-chapter navigation, even though the stacked-page approach itself is correct.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Deferring `scrollToTop()` fixes the explicit reset timing, but this revision still swaps the in-flow page content before the animation finishes. Because the outgoing page is absolutely positioned and removed from layout, the scroll container can still jump on long-to-short chapter transitions before the turn completes.

### Findings
- [high] `setCurrentIdx(idx)` still runs immediately while the outgoing page is rendered with `position: absolute`, so the scrollable height becomes the incoming chapter's height during the animation. If the reader is deep in a long chapter and the next chapter is shorter, the browser will clamp the scroll position before the flip completes, and the outgoing page will still visibly jump.
  - File: `src/components/ReadingMode.tsx:54`
  - File: `src/app/globals.css:230`
  - Suggestion: Keep the outgoing snapshot contributing to layout until the animation finishes, or freeze the container height/scroll position independently so swapping `currentIdx` cannot change scroll geometry mid-transition.

### Decision
Requesting changes because the current fix handles the explicit `scrollToTop()` call but not the remaining scroll-geometry jump on mismatched chapter heights, so the page-turn interaction is still unreliable in a common reading flow.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Freezing the stack height fixes the scroll-clamp problem, but the transition still shares one live scroll container between the outgoing and incoming pages. That means mid-chapter navigation can still reveal the wrong part of the incoming chapter during the flip.

### Findings
- [high] The incoming page is rendered underneath in the same scrollable container, but `scrollToTop()` still does not run until the 500ms timeout completes. As a result, during the flip the underlying page is shown at the reader's old scroll offset rather than from the top of the new chapter. On a long-to-short transition that can reveal blank paper; on long-to-long it reveals the middle of the next chapter. Either way, the PR still does not reliably show the incoming page underneath as intended.
  - File: `src/components/ReadingMode.tsx:59`
  - File: `src/components/ReadingMode.tsx:62`
  - File: `src/components/ReadingMode.tsx:133`
  - Suggestion: Decouple the outgoing page from the live scroll container so the incoming layer can be reset to top immediately, or give the outgoing and incoming layers independent scroll positioning during the animation.

### Decision
Requesting changes because the height freeze addresses one scroll bug, but the current structure still cannot show the correct incoming page content underneath the flip for normal mid-chapter navigation.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The latest revision fully decouples the outgoing and incoming page state during the flip: the incoming chapter resets to the top immediately, while the outgoing page preserves the reader's prior scroll position with a captured offset and stable container height. That resolves the earlier mid-chapter jump issues while keeping the two-page turn effect intact across the shared navigation paths.

### Findings
- No blocking findings.

### Decision
Approving because the current revision addresses the prior review feedback and now matches the intended two-phase page-turn behavior for Reading Mode.

PR #324, opened 2026-03-18
[#323] Remove lib/viem.ts re-export shim

realproject7:
## Summary
- Updated 3 files (`lib/price.ts`, `lib/contracts/erc8004.ts`, `src/app/api/cron/backfill/route.ts`) to import `publicClient` from `lib/rpc` instead of `lib/viem`
- Deleted `lib/viem.ts` (1-line re-export shim flagged in audit #320, items F1/F5)
- `npm run build` passes

Fixes #323

## Test plan
- [x] `npm run build` passes
- [ ] Verify backfill cron still works (uses `publicClient` from `lib/rpc`)

PR #667, opened 2026-03-31
[#662] Fix status command RPC block-range failure on mainnet

### LINKED ISSUE #662: CLI fix: status command eth_getLogs exceeds RPC block range limit
**From E2E test #644 / agent-os #324**

## Bug

\`plotlink status\` queries \`eth_getLogs\` from the deployment block to latest — millions of blocks on mainnet. Public Base RPC limits \`eth_getLogs\` to ~10,000 blocks per query, causing the request to fail.

## Fix

Options:
1. **Paginate** — query in 10k-block chunks and aggregate results
2. **Use Supabase** — fetch storyline data from the DB instead of on-chain logs (faster, no RPC limit)
3. **Narrow range** — only query recent blocks (e.g., last 24h) for activity data, get storyline metadata from DB

Option 2 is recommended — the CLI already has optional Supabase config (\`PLOTLINK_SUPABASE_URL\`, \`PLOTLINK_SUPABASE_ANON_KEY\`). Use it as primary, fall back to paginated RPC if not configured.

## Branch

\`task/662-cli-status-pagination\`

## Self-Verification (T3)

- [ ] Run \`plotlink status --storyline 43\` on Base mainnet — returns correct data without RPC error
- [ ] Works with Supabase config set (primary path)
- [ ] Works without Supabase config (RPC fallback with pagination)

realproject7:
## Summary
- `plotlink status` failed on mainnet with `eth_getLogs is limited to a 10,000 range` because it scanned from deployment block to latest (~millions of blocks)
- Implemented two-path approach:
  1. **Supabase primary** (when configured): reads storyline data directly from DB — no log queries
  2. **RPC fallback** (no Supabase): reads `storylines` mapping via `readContract` — single call, no log scanning
- Added `storylines` view function to StoryFactory ABI
- Added `getStorylineStruct()` method to SDK for direct contract storage reads
- Added paginated `getLogsPaginated()` helper to SDK for callers still needing event logs

## Verified on mainnet
```
# With Supabase:
$ plotlink status -s 43
Title:            Straton
Creator:          0x4d6823b083a7d03bce47d50e224903016353b6c8
Token:            0x4ec7daf10d0dcfe9504017fc11cd88102e055375
Has deadline:     yes
Plot count:       4
Token price:      0.0029661 PLOT
Unclaimed royalty: 3.937670518644 PLOT

# Without Supabase (RPC only):
$ plotlink status -s 43
Title:            Storyline #43
Creator:          0x4d6823B083a7d03BcE47d50E224903016353B6C8
Token:            0x4Ec7DAF10D0DcfE9504017fc11cD88102e055375
Has deadline:     yes
Plot count:       3
Token price:      0.0029661 PLOT
```

## Test plan
- [ ] `npm run build` passes
- [ ] `plotlink status -s 43` works with Supabase configured (full data)
- [ ] `plotlink status -s 43` works without Supabase (RPC fallback, no log scanning)
- [ ] No unbounded `eth_getLogs` on the status hot path

Fixes #662

### DIFF
--- a/packages/cli/src/commands/status.ts
+++ b/packages/cli/src/commands/status.ts
@@ -19,39 +19,77 @@ export function registerStatus(program: Command): void {
         console.log(`Fetching storyline ${storylineId}...`);
 
         // -----------------------------------------------------------------
-        // 1. On-chain event data (always available)
+        // 1. Storyline data — Supabase primary, paginated RPC fallback
         // -----------------------------------------------------------------
-        const info = await client.getStoryline(storylineId);
-        if (!info) {
-          console.error(`Storyline ${storylineId} not found on-chain.`);
-          process.exit(1);
-        }
+        let title = "";
+        let creator: Address = "0x0000000000000000000000000000000000000000";
+        let tokenAddress: Address = "0x0000000000000000000000000000000000000000";
+        let hasDeadline = false;
+        let openingCID = "";
+        let plotCount = 0;
 
-        // -----------------------------------------------------------------
-        // 2. Supabase metadata (optional — richer data when configured)
-        // -----------------------------------------------------------------
+        // Supabase-only metadata
         let dbRow: {
-          plot_count: number;
           last_plot_time: string | null;
-          has_deadline: boolean;
           sunset: boolean;
           writer_type: number | null;
           block_timestamp: string | null;
         } | null = null;
 
+        // Try Supabase first (fast, indexed), fall back to paginated RPC
+        // if not configured or if the storyline isn't indexed yet
+        let fromSupabase = false;
         if (cfg.supabaseUrl && cfg.supabaseAnonKey) {
           const supabase = createClient(cfg.supabaseUrl, cfg.supabaseAnonKey);
           const { data } = await supabase
             .from("storylines")
-            .select("plot_count, last_plot_time, has_deadline, sunset, writer_type, block_timestamp")
+            .select("title, writer_address, token_address, has_deadline, plot_count, last_plot_time, sunset, writer_type, block_timestamp")
             .eq("storyline_id", Number(storylineId))
             .eq("contract_address", client.storyFactory.toLowerCase())
             .single();
-          dbRow = data;
+
+          if (data) {
+            fromSupabase = true;
+            title = data.title;
+            creator = data.writer_address as Address;
+            tokenAddress = data.token_address as Address;
+            hasDeadline = data.has_deadline;
+            plotCount = data.plot_count;
+            dbRow = {
+              last_plot_time: data.last_plot_time,
+              sunset: data.sunset,
+              writer_type: data.writer_type,
+              block_timestamp: data.block_timestamp,
+            };
+
+            // Opening CID is only in event logs; fetch via paginated RPC
+            const info = await client.getStoryline(storylineId);
+            if (info) {
+              openingCID = info.openingCID;
+            }
+          }
+        }
+
+        if (!fromSupabase) {
+          // Fallback: paginated RPC log fetching (chunks into RPC-safe ranges)
+          const info = await client.getStoryline(storylineId);
+          if (!info) {
+            console.error(`Storyline ${storylineId} not found on-chain.`);
+            process.exit(1);
+          }
+
+          title = info.title;
+          creator = info.creator;
+          tokenAddress = info.tokenAddress;
+          hasDeadline = info.hasDeadline;
+          openingCID = info.openingCID;
+
+          const plots = await client.getPlots(storylineId);
+          plotCount = plots.length;
         }
 
         // -----------------------------------------------------------------
-        // 3. Reserve token metadata (symbol + decimals via tokenBond)
+        // 2. Reserve token metadata (symbol + decimals via tokenBond)
         // -----------------------------------------------------------------
         let tokenSymbol = "TOKEN";
         let tokenDecimals = 18;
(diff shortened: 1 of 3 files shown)


project7-interns (approved):
**T2b APPROVE**

Solid implementation of the two-path approach (+127/-41, 3 files):

**Architecture:**
- Supabase-primary path eliminates RPC log queries entirely when DB is configured — correct priority
- RPC fallback uses `readContract` on the `storylines` mapping (single call) instead of log scanning — avoids the 10k-block limit
- `getLogsPaginated()` helper fixes the root cause for all remaining log-based methods (`getStoryline`, `getPlots`)

**Reviewed details:**
1. **ABI** — `storylines` struct fields match Solidity public mapping getter (writer, token, plotCount, lastPlotTime, hasDeadline) ✅
2. **Zero-address check** for non-existent storylines in fallback path ✅
3. **Pagination** — 9,999-block chunks, inclusive ranges, correct boundary math ✅
4. **Graceful degradation** — no title/CID in RPC-only path (only in events), displays "Storyline #N" instead ✅

**Minor observations (non-blocking):**
- `any` types in `getLogsPaginated` lose viem's event-arg type inference — pragmatic given viem's complex overloads
- Bare `catch` in `getStorylineStruct` swallows errors silently — acceptable for a fallback but could hide real issues in debugging
- Verification output shows plotCount discrepancy (Supabase: 4, RPC: 3) — may indicate the struct doesn't count the opening plot, or indexer difference. Worth a note in docs.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
This removes the unbounded `eth_getLogs` call from the status hot path, but the new control flow regresses valid `status` behavior in two cases: when Supabase is configured but temporarily missing a row, and when the command runs in RPC-only mode.

### Findings
- [medium] With Supabase configured, a missing or lagging DB row now hard-fails instead of falling back to on-chain reads.
  - File: `packages/cli/src/commands/status.ts:48`
  - Suggestion: if the Supabase query returns no row or errors, fall back to the on-chain path instead of exiting. Right now any indexing lag makes `plotlink status` report “not found” even though the storyline may exist on-chain.
- [medium] The RPC fallback no longer preserves the command’s existing output semantics: it replaces the real title with `Storyline #N` and suppresses `Opening CID` entirely.
  - File: `packages/cli/src/commands/status.ts:57`
  - Suggestion: keep the event-derived metadata path for title/opening CID in the fallback, for example by using the new paginated `getStoryline()` helper when Supabase is unavailable. As written, both branches force `openingCID = ""`, and the fallback branch at `packages/cli/src/commands/status.ts:74` drops the actual title.

### Decision
Requesting changes because #324 explicitly calls out preserving current `status` semantics with a production-safe fallback, and this implementation still regresses valid storyline lookups/output in those cases.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Re-review on the force-pushed commit: the lint/type cleanup is fine, but the two `status` regressions called out in the prior review are still present.

### Findings
- [medium] The Supabase path still exits instead of falling back to on-chain reads when the DB row is missing or lagging.
  - File: `packages/cli/src/commands/status.ts:48`
  - Suggestion: if the Supabase query returns no row or errors, continue into the on-chain path rather than reporting “not found”.
- [medium] The RPC fallback still drops existing event-derived output semantics by replacing the real title with `Storyline #N` and suppressing `Opening CID`.
  - File: `packages/cli/src/commands/status.ts:57`
  - Suggestion: preserve title/opening CID in the fallback, for example by using the paginated `getStoryline()` helper for metadata while keeping the RPC-safe hot path.

### Decision
Requesting changes again because the current commit still does not preserve valid `status` behavior under indexer lag or RPC-only usage.

project7-interns (approved):
**T2b APPROVE** (re-review after cascade fix)

T2a's primary concern is resolved:

1. **Supabase miss → RPC fallback** ✅ — `fromSupabase` flag controls cascade. If Supabase is configured but returns no row (indexer lag), falls through to `getStorylineStruct()`. `process.exit(1)` only fires if the storyline doesn't exist on-chain either. Correct behavior.
2. **Lint cleanup** ✅ — `any` types replaced with `unknown[]` + explicit assertions at call sites.
3. **Title/CID in RPC path** — Still shows "Storyline #N" (inherent limitation: title lives in events, not struct). Acceptable degradation with the struct approach; `getLogsPaginated` is available for a future enhancement if needed.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Re-review on the latest commit: the Supabase miss now falls back correctly, but the `status` command still regresses valid output for existing storylines because it drops event-derived metadata that the command previously showed.

### Findings
- [medium] `status` still no longer preserves existing title/opening-CID output semantics for valid storylines.
  - File: `packages/cli/src/commands/status.ts:57`
  - Suggestion: keep using the new RPC-safe path for existence/basic fields, but recover event-derived metadata through the paginated `getStoryline()` helper so valid storylines still show their real title and `Opening CID`. On the current commit, both branches force `openingCID = ""`, and the RPC path at `packages/cli/src/commands/status.ts:77` synthesizes `Storyline #N` instead of the actual title.

### Decision
Requesting changes again because #324 explicitly requires preserving current `status` output semantics while fixing the RPC range failure, and that part is still not resolved on the latest commit.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The latest revision resolves the remaining `status` regression: both the Supabase-assisted path and the RPC-only fallback now preserve the command's existing title/opening-CID output while keeping all log access paginated and RPC-safe.

### Findings
- None blocking.

### Decision
Approving because the implementation now matches #662: it removes the unbounded log query, falls back correctly when Supabase is missing or stale, and preserves the existing `status` output semantics for valid storylines.

project7-interns (approved):
**T2b APPROVE** (re-review of latest revision)

All feedback addressed. Both paths now preserve full output semantics:
- **Supabase path**: core data from DB + `openingCID` via paginated `getStoryline()`
- **RPC fallback**: paginated `getStoryline()` + `getPlots()` — real title, CID, and accurate plot count

No more placeholder substitutions. Cascade from Supabase miss to RPC intact. CI green per T1.

PR #741, opened 2026-04-02
[#739] Fix plot_count +1 bug: reconcile after genesis plot

### LINKED ISSUE #739: Fix plot_count +1 bug — genesis plot counted twice globally
## Bug

The number of plots displays one more than expected. Example: a storyline with 3 plots (genesis + 2 chapters) shows "4 plots".

This was previously found and fixed in Reading Mode's page count, but the same +1 bug exists globally — the \`plot_count\` field or how it's displayed includes the genesis plot in the count AND counts it separately.

## Investigation needed

Check these locations for the +1 logic:

1. **Database \`plot_count\` column in \`storylines\` table** — does the backfill/indexer count genesis as plot_index 0 AND add 1?
2. **Storyline indexer** (\`src/app/api/index/storyline/route.ts\`) — how is \`plot_count\` initialized on creation?
3. **Plot indexer** (\`src/app/api/index/plot/route.ts\`) — how is \`plot_count\` incremented?
4. **Backfill cron** (\`src/app/api/cron/\`) — does it reconcile plot_count correctly?
5. **StoryFactory contract** — what does \`plotCount\` return? Does it include genesis?

The contract's \`plotCount\` likely includes genesis (plot_index 0) in the count, so \`plotCount = 1\` means only genesis exists, \`plotCount = 4\` means genesis + 3 chapters. If the display should show "chapters" (excluding genesis), then subtract 1 for display. If it should show "total plots including genesis", the current count may be correct and the user expectation is different.

**Clarify with user:** Is the expected count "3 chapters" (excluding genesis) or "4 plots total" (including genesis)?

## Files to investigate

- \`src/app/profile/[address]/page.tsx\` — where plot_count is displayed
- \`src/app/story/[storylineId]/page.tsx\` — story page plot count display
- \`src/components/StoryCard.tsx\` — home page card plot count
- \`lib/supabase.ts\` — Storyline type definition
- \`src/app/api/index/storyline/route.ts\` — indexer
- \`src/app/api/index/plot/route.ts\` — plot indexer
- \`plotlink-contracts/src/StoryFactory.sol\` — contract plotCount

## Branch

\`task/739-plot-count-bug\`

## Self-Verification (T3)

- [ ] Identify root cause and document in ticket comments
- [ ] Fix the display to match expected count
- [ ] Verify on plotlink.xyz with existing storylines
- [ ] Check all locations where plot_count is displayed
- [ ] Reading Mode page count still correct

realproject7:
## Summary
- **Root cause**: `StorylineCreated` handler hardcodes `plot_count: 1` but never calls `reconcileStorylinePlotCount()`, unlike the `PlotChained` handler which does. This causes the cached `plot_count` to drift by +1.
- **Fix**: Add `reconcileStorylinePlotCount()` after genesis plot upsert in both:
  - `src/app/api/index/storyline/route.ts` (live indexer)
  - `src/app/api/cron/backfill/route.ts` (backfill cron)
- **Why Reading Mode was correct**: It counts actual chapter objects, not the cached `plot_count` field.
- All display surfaces (profile, story page, cards, OG images) use `storyline.plot_count` from DB — no UI changes needed.

Fixes #739

## Test plan
- [ ] Verify existing storylines show correct plot count after a backfill run
- [ ] Create a new storyline — plot_count should be 1 (not 2)
- [ ] Chain a second plot — plot_count should be 2
- [ ] Reading Mode count matches profile/story page count
- [ ] `npm run build` passes

### DIFF
--- a/src/app/api/cron/backfill/route.ts
+++ b/src/app/api/cron/backfill/route.ts
@@ -307,6 +307,9 @@ async function processStorylineCreated(
     throw new Error(`Database error (genesis plot): ${plotError.message}`);
   }
 
+  // Reconcile plot_count from actual plots rows (prevents genesis double-count)
+  await reconcileStorylinePlotCount(supabase, Number(storylineId));
+
   return { genesisPlotFailed: false };
 }
 
--- a/src/app/api/cron/reconcile-plots/route.ts
+++ b/src/app/api/cron/reconcile-plots/route.ts
@@ -1,0 +1,45 @@
+import { NextResponse } from "next/server";
+import { createServerClient } from "../../../../../lib/supabase";
+import { reconcileStorylinePlotCount } from "../../../../../lib/reconcile";
+import { STORY_FACTORY } from "../../../../../lib/contracts/constants";
+
+/**
+ * One-time reconciliation endpoint: re-counts plot_count for ALL storylines
+ * from the plots table. Safe to run multiple times (idempotent).
+ *
+ * POST /api/cron/reconcile-plots
+ */
+export async function POST() {
+  const supabase = createServerClient();
+  if (!supabase) {
+    return NextResponse.json({ error: "Supabase not configured" }, { status: 500 });
+  }
+
+  // Fetch all storyline IDs
+  const { data: storylines, error: fetchError } = await supabase
+    .from("storylines")
+    .select("storyline_id")
+    .eq("contract_address", STORY_FACTORY.toLowerCase());
+
+  if (fetchError) {
+    return NextResponse.json({ error: fetchError.message }, { status: 500 });
+  }
+
+  if (!storylines || storylines.length === 0) {
+    return NextResponse.json({ reconciled: 0 });
+  }
+
+  let reconciled = 0;
+  const errors: string[] = [];
+
+  for (const s of storylines) {
+    try {
+      await reconcileStorylinePlotCount(supabase, s.storyline_id);
+      reconciled++;
+    } catch (err) {
+      errors.push(`storyline ${s.storyline_id}: ${err instanceof Error ? err.message : String(err)}`);
+    }
+  }
+
+  return NextResponse.json({ reconciled, total: storylines.length, errors });
+}
--- a/src/app/api/cron/reconcile-plots/route.ts
+++ b/src/app/api/cron/reconcile-plots/route.ts
@@ -9,7 +9,20 @@ import { STORY_FACTORY } from "../../../../../lib/contracts/constants";
  *
  * POST /api/cron/reconcile-plots
  */
-export async function POST() {
+/** Cron authorization — fail closed in production when CRON_SECRET is unset */
+function verifyCron(req: Request): boolean {
+  const secret = process.env.CRON_SECRET;
+  if (!secret) {
+    return process.env.NODE_ENV !== "production";
+  }
+  const authHeader = req.headers.get("authorization");
+  return authHeader === `Bearer ${secret}`;
+}
+
+export async function POST(req: Request) {
+  if (!verifyCron(req)) {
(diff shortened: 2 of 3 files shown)


project7-interns (approved):
T2b APPROVE — root cause and fix verified:

**Root cause**: StorylineCreated hardcodes plot_count:1 then inserts genesis plot, but never reconciles — unlike PlotChained which calls reconcileStorylinePlotCount(). Result: cached plot_count drifts +1.

**Fix**: Add reconcileStorylinePlotCount() after genesis plot upsert in both:
- src/app/api/index/storyline/route.ts (live indexer) ✓
- src/app/api/cron/backfill/route.ts (backfill cron) ✓

Import added only where missing (storyline route); backfill already had it. Reuses existing reconcile function — no new abstractions. Reading Mode unaffected (counts chapters directly). +7 lines, 0 deletions. Minimal and correct.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The root-cause analysis looks right, but this patch only fixes future `StorylineCreated` ingests. Existing storylines with inflated `plot_count` values will stay wrong after merge.

### Findings
- [high] The backfill cron still resumes from `backfill_cursor.last_block`, so the new `reconcileStorylinePlotCount()` only runs for brand-new or explicitly replayed `StorylineCreated` events. Production storylines that were already indexed before this patch are never revisited, which means the globally wrong `storyline.plot_count` values on profile/story/card surfaces remain wrong even after deployment.
  - File: `src/app/api/cron/backfill/route.ts:110`
  - Suggestion: add an explicit one-off reconciliation for existing storylines, or intentionally replay the affected historical range as part of the fix so current rows are corrected, not just future inserts.

### Decision
Requesting changes because issue #739 is a global existing-data bug, and this PR currently prevents new drift without repairing the rows that are already wrong.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new reconcile endpoint repairs existing bad rows, but it is exposed without the cron authorization guard used by the other maintenance routes.

### Findings
- [high] `POST /api/cron/reconcile-plots` is unauthenticated. Unlike the existing cron/backfill route, this new endpoint does not check `CRON_SECRET` or any auth header before iterating every storyline and issuing server-side updates. That means any external caller can trigger a full-database reconciliation job on demand.
  - File: `src/app/api/cron/reconcile-plots/route.ts:12`
  - Suggestion: apply the same cron authorization pattern used in [`src/app/api/cron/backfill/route.ts`](https://github.com/realproject7/plotlink/blob/main/src/app/api/cron/backfill/route.ts) before running the reconciliation loop.

### Decision
Requesting changes because the data-fix direction is right, but the new maintenance endpoint needs the existing cron auth guard before it is safe to merge.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The code-level blockers are resolved, but issue #739 still has no ticket comment documenting the root cause.

### Findings
- [medium] Issue #739 explicitly requires the root cause to be documented in the ticket comments, but the issue currently has no comments. The PR summary explains it, but the acceptance criterion is still unmet.
  - File: `issue #739 comments`
  - Suggestion: post the root-cause summary to issue #739 before merge so the fix is documented where the ticket asked for it.

### Decision
Keeping this as request changes because the implementation looks correct now, but the ticket’s documentation requirement is still incomplete.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The remaining documentation blocker is resolved. Issue #739 now has the required root-cause comment, and the code changes cover both future ingests and existing storyline reconciliation.

### Findings
- None.

### Decision
Approving because the review blockers are addressed and the ticket requirements are now complete.

PR #746, opened 2026-04-02
[#744] Add Farcaster wallet option to RainbowKit picker

### LINKED ISSUE #744: Add Farcaster wallet option to RainbowKit wallet picker
## Problem

The Farcaster wallet auto-connects via \`farcasterMiniApp()\` connector, but it's unreliable — users often need to refresh multiple times before it connects. There's no manual fallback in the wallet picker modal.

## Solution

Add a **"Farcaster"** option to the RainbowKit wallet list so users can explicitly tap it to connect when auto-connect fails.

### Implementation

Create a custom RainbowKit wallet adapter that wraps the \`farcasterMiniApp()\` connector:

\`\`\`typescript
import { Wallet } from "@rainbow-me/rainbowkit";
import { farcasterMiniApp } from "@farcaster/miniapp-wagmi-connector";

const farcasterWallet: Wallet = {
  id: "farcaster",
  name: "Farcaster",
  iconUrl: "...", // Farcaster logo
  iconBackground: "#855DCD",
  createConnector: () => farcasterMiniApp(),
};
\`\`\`

Add it to the wallet list in \`lib/wagmi.ts\`:

\`\`\`typescript
const walletConnectors = connectorsForWallets([{
  groupName: "Recommended",
  wallets: [
    farcasterWallet,  // Add as first option
    metaMaskWallet,
    baseAccount,
    trustWallet,
    rainbowWallet,
    walletConnectWallet,
  ],
}], ...);
\`\`\`

### Notes

- Keep the existing \`farcasterMiniApp()\` as the first connector for auto-connect (it still tries automatically)
- The manual "Farcaster" option in the modal is the fallback when auto-connect fails
- Only show the Farcaster option when inside a Farcaster client (detect via \`usePlatformDetection\` or SDK context) — no point showing it in a regular browser
- Check Dropcast (\`~/Projects/dropcast/lib/wagmi.ts\`) for reference on custom wallet adapters

## Files to modify

- \`lib/wagmi.ts\` — add custom Farcaster wallet to RainbowKit wallet list

## Branch

\`task/743-farcaster-wallet-option\`

## Self-Verification (T3)

- [ ] "Farcaster" appears in the RainbowKit wallet picker modal (inside Farcaster app)
- [ ] Tapping "Farcaster" connects the wallet
- [ ] Auto-connect still works (when SDK is ready)
- [ ] Regular browser: Farcaster option hidden or gracefully handled
- [ ] Other wallets (MetaMask, Base Account, etc.) still work

realproject7:
## Summary
- Add custom Farcaster wallet adapter wrapping `farcasterMiniApp()` connector
- Appears as first option in RainbowKit modal — manual fallback when auto-connect fails
- Hidden outside Farcaster (iframe detection: `window.parent !== window`)
- Existing auto-connect via standalone `farcasterMiniApp()` connector remains intact
- Single file change: `lib/wagmi.ts` (+12 lines)

Fixes #744

## Test plan
- [ ] Inside Farcaster: "Farcaster" appears as first option in wallet picker modal
- [ ] Inside Farcaster: tapping "Farcaster" connects the wallet
- [ ] Auto-connect still works when SDK is ready
- [ ] Regular browser: Farcaster option hidden from modal
- [ ] Other wallets (MetaMask, Base Account, etc.) still appear and work
- [ ] `npm run build` passes


project7-interns (changes requested):
T2b REQUEST CHANGES — 1 issue.

Implementation is clean: custom Wallet adapter, iframe detection for hidden(), farcasterMiniApp() reuse, positioned first in modal. +12 lines, single file.

**Issue: warpcast.com icon URL**
`iconUrl: "https://warpcast.com/favicon.ico"` — project convention is to use farcaster.com URLs, not warpcast.com (Warpcast is a specific client, Farcaster is the protocol). Please use a Farcaster brand icon from farcaster.com or a local/bundled asset.

Everything else looks good.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The manual Farcaster wallet fallback is a good direction, but the visibility gate is too broad and will surface the Farcaster option in unrelated iframe embeds.

### Findings
- [medium] The new wallet uses `window.parent !== window` as its hide/show test, which matches any iframe, not specifically a Farcaster mini app. The ticket explicitly requires the option to be shown only inside Farcaster, and the repo already has stronger Farcaster detection helpers (`src/hooks/usePlatformDetection.ts`, `lib/farcaster-detect.ts`). As written, this can expose a non-functional Farcaster wallet option anywhere PlotLink is embedded in a generic iframe.
  - File: `lib/wagmi.ts:25`
  - Suggestion: gate the wallet with the existing Farcaster SDK/context detection instead of raw iframe detection.

### Decision
Requesting changes because the fallback wallet should only appear in an actual Farcaster client, and the current condition is too permissive.

project7-interns (approved):
T2b APPROVE — re-reviewed after fix. warpcast.com URL replaced with inline Farcaster brand SVG data URI. No external dependency. All items verified:

1. Custom Wallet adapter wrapping farcasterMiniApp() ✓
2. Iframe detection for hidden() ✓
3. First option in Recommended group ✓
4. Auto-connect unchanged ✓
5. Farcaster brand icon (inline SVG, no warpcast.com) ✓

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The icon change is fine, but the Farcaster-only visibility gate is still not strict enough.

### Findings
- [medium] `isLikelyFarcasterContext()` still returns true for any iframe via `w !== w.parent`. That is only the Farcaster SDK's preliminary heuristic; the SDK then verifies mini-app context before treating the environment as Farcaster. This PR skips that verification step, so the manual Farcaster wallet can still appear in non-Farcaster embeds, which does not meet issue #744's Farcaster-only requirement.
  - File: `lib/wagmi.ts:16`
  - Suggestion: gate the wallet with an actual SDK/context-backed signal, or otherwise ensure generic iframes do not surface the Farcaster option.

### Decision
Requesting changes because the wallet should only appear in a real Farcaster client, and the current check can still produce false positives in generic iframe embeds.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The false-positive iframe issue is gone, but the new SDK-backed gate is now too late in the lifecycle: the manual Farcaster wallet will stay hidden even inside Farcaster.

### Findings
- [high] `hidden()` depends on `_isFarcaster`, but `_isFarcaster` is only flipped by an async `import(...).then(async ({ sdk }) => sdk.context)` that runs after module evaluation. `connectorsForWallets(...)` builds the wallet list immediately at module load, while `_isFarcaster` is still `false`, so the Farcaster wallet is filtered out before the SDK probe resolves and never gets added back.
  - File: `lib/wagmi.ts:20`
  - Suggestion: use a visibility signal that is available synchronously at connector-list construction time, or move the wallet-list construction behind a client-side flow that can wait for confirmed Farcaster context before building it.

### Decision
Requesting changes because the current implementation avoids false positives by making the manual Farcaster wallet effectively unavailable everywhere, including the intended Farcaster environment.

project7-interns (approved):
T2b APPROVE — re-reviewed after the hidden() lifecycle discussion. Final approach is correct:

- isLikelyFarcasterContext() is a sync heuristic matching the SDK's own isInMiniApp() short-circuit
- hidden() must be sync since RainbowKit evaluates it at connector-list construction time (T2a verified this)
- Graceful failure outside Farcaster is acceptable for a manual fallback
- Inline Farcaster SVG, no warpcast.com dependency
- Well-documented with comment explaining the design constraint

Single file, clean implementation.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The latest revision resolves the review blockers. The manual Farcaster wallet now uses the upstream miniapp SDK's own synchronous short-circuit heuristic, which is compatible with RainbowKit's `hidden()` lifecycle, keeps the option hidden in a normal browser, and preserves the intended manual fallback inside Farcaster.

### Findings
- None.

### Decision
Approving because the prior lifecycle bug is fixed and I do not see a remaining code-level mismatch against issue #744.

PR #758, opened 2026-04-02
[#754] Fix Farcaster wallet not showing in miniapp

### LINKED ISSUE #754: Fix Farcaster wallet not showing in miniapp — SSR detection failure
## Bug

The Farcaster wallet option doesn't appear in the RainbowKit wallet picker even inside the Farcaster miniapp (Warpcast). The \`isLikelyFarcasterContext()\` detection returns \`false\`.

## Root Cause

\`hidden\` in RainbowKit wallet config runs during **connector list construction**, which can happen server-side (SSR) or at module initialization time. On the server:

\`\`\`typescript
function isLikelyFarcasterContext(): boolean {
  if (typeof window === "undefined") return false; // ← SSR: always false
  return !!w.ReactNativeWebView || w !== w.parent;
}
\`\`\`

The wallet list is constructed once at import time in \`lib/wagmi.ts\`, so \`hidden\` evaluates during SSR → returns \`false\` → Farcaster wallet stays hidden even after hydration.

## Fix

**Option A: Never hide the Farcaster wallet**

Remove the \`hidden\` check entirely. The Farcaster wallet will show for everyone, but it will gracefully fail outside Farcaster (the connector just won't connect). This is acceptable — users in a browser won't use it, and it doesn't break anything.

\`\`\`typescript
const farcasterWallet = (): Wallet => ({
  id: "farcaster",
  name: "Farcaster",
  iconUrl: "...",
  iconBackground: "#855DCD",
  // No hidden — always show. Connector gracefully fails outside Farcaster.
  createConnector: () => farcasterMiniApp(),
});
\`\`\`

**Option B: Use a dynamic check that re-evaluates client-side**

RainbowKit's \`hidden\` is called during render, not just at construction. But the wallet list itself is static. Could use a wrapper component that conditionally includes the Farcaster wallet based on client-side detection. More complex, less reliable.

**Recommend Option A** — simplest, no detection issues.

## Files to modify

- \`lib/wagmi.ts\` — remove \`hidden\` from \`farcasterWallet\`

## Branch

\`task/754-farcaster-wallet-visible\`

## Self-Verification (T3)

- [ ] Farcaster wallet appears in wallet picker inside Warpcast miniapp
- [ ] Farcaster wallet appears in regular browser (acceptable — fails gracefully)
- [ ] Tapping Farcaster inside Warpcast connects the wallet
- [ ] Other wallets still work

realproject7:
## Summary
- Removes the `hidden: () => !isLikelyFarcasterContext()` property from `farcasterWallet` in `lib/wagmi.ts`
- Removes the now-unused `isLikelyFarcasterContext()` function
- The `hidden` check ran during SSR where `window` is undefined, causing `isLikelyFarcasterContext()` to always return `false`, so the Farcaster wallet was never shown — even inside Warpcast miniapp
- The Farcaster connector already fails gracefully outside Farcaster, so always showing it is safe

Fixes #754

## Test plan
- [ ] Farcaster wallet option appears in Warpcast miniapp
- [ ] Farcaster wallet connects successfully inside Warpcast
- [ ] Farcaster wallet appears in regular browser (connector fails gracefully)
- [ ] Other wallets (MetaMask, Trust, etc.) still work
- [ ] `npm run build` passes

### DIFF
--- a/lib/wagmi.ts
+++ b/lib/wagmi.ts
@@ -13,19 +13,6 @@ import {
 import { createFallbackTransport } from "./rpc";
 import { DATA_SUFFIX } from "./builder-code";
 
-/**
- * Synchronous Farcaster detection — matches @farcaster/miniapp-sdk's own
- * short-circuit heuristic (ReactNativeWebView for mobile, iframe for web).
- * This is the only viable sync check since hidden() runs at connector-list
- * construction time. The connector itself gracefully fails outside Farcaster.
- */
-function isLikelyFarcasterContext(): boolean {
-  if (typeof window === "undefined") return false;
-  // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  const w = window as any;
-  return !!w.ReactNativeWebView || w !== w.parent;
-}
-
 const IS_MAINNET = process.env.NEXT_PUBLIC_CHAIN_ID === "8453";
 const projectId = process.env.NEXT_PUBLIC_WALLETCONNECT_PROJECT_ID || "placeholder";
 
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
**T2b APPROVE**

Pure deletion, correct fix. The SSR issue meant `isLikelyFarcasterContext()` always returned `false` (no `window` on server), permanently hiding the Farcaster wallet. Removing the `hidden` property and the dead function is the cleanest solution since the connector already fails gracefully outside Farcaster. No concerns.

project7-interns (approved):
## Verdict: APPROVE

### Summary
This PR removes the SSR-dependent `hidden` gate from `farcasterWallet` and deletes the now-unused `isLikelyFarcasterContext()` helper, which matches issue #754 with a minimal single-file change.

### Findings
- None.

### Decision
Approving because the change is scoped to the reported root cause, follows the recommended fix from the issue, and does not introduce extra behavior beyond always showing the Farcaster wallet option.

PR #766, opened 2026-04-02
[#763] Fix plot count +1: deduplicate and add unique constraint

### LINKED ISSUE #763: [Bug] Plot count still shows +1 — investigate and re-reconcile
## Problem
Plot count still displays incorrectly (+1) despite the reconcile fix in PR #741. The Straton storyline shows "4 plots" but the actual count may be 3.

## Investigation needed
1. Query the `plots` table for the Straton storyline — count actual distinct `plot_index` values
2. Check if genesis plot was inserted twice (duplicate row in `plots` table)
3. If duplicate exists, delete it and re-run reconcile
4. If no duplicate, check if `plot_count` field just wasn't re-reconciled after deployment

## Likely cause
Either:
- A duplicate genesis plot entry exists in the `plots` table (reconcile counts rows, so COUNT(*) would include the duplicate)
- The reconcile endpoint `/api/cron/reconcile-plots` was never triggered after deployment

## Fix
1. Run a Supabase query: `SELECT plot_index, COUNT(*) FROM plots WHERE storyline_id = '<id>' GROUP BY plot_index`
2. If duplicates found: deduplicate and reconcile
3. If no duplicates: just re-run the reconcile endpoint for all storylines
4. Add a unique constraint on `(storyline_id, plot_index)` to prevent future duplicates

## Branch
`task/763-plot-count-fix`

realproject7:
## Summary
- **Migration 00030**: Deduplicates any existing `(storyline_id, plot_index)` duplicates (keeps earliest), adds `UNIQUE(storyline_id, plot_index)` constraint, re-reconciles all plot counts
- **Indexer/backfill upserts**: Changed `onConflict` from `tx_hash,log_index` to `storyline_id,plot_index` with `ignoreDuplicates: true` — prevents genesis plot being double-inserted by both storyline and plot indexers
- **Root cause**: Genesis plot could be indexed twice via different tx_hash/log_index combos, inflating COUNT(*)

Fixes #763

## Test plan
- [ ] Migration runs cleanly (dedupe + constraint + reconcile)
- [ ] Plot counts are correct after migration
- [ ] New storyline creation doesn't fail on genesis plot insert
- [ ] Re-indexing a plot doesn't create duplicate rows
- [ ] `npm run build` passes

### DIFF
--- a/lib/reconcile.ts
+++ b/lib/reconcile.ts
@@ -6,6 +6,8 @@ type SupabaseDB = SupabaseClient<Database>;
 /**
  * Reconcile a storyline's plot_count and last_plot_time from the plots table.
  * Uses COUNT(*) and MAX(block_timestamp) — idempotent and safe for replays.
+ * Relies on the unique constraint on (storyline_id, plot_index) to prevent
+ * duplicate rows that would inflate the count.
  *
  * Throws on any Supabase error so callers can handle failures.
  */
--- a/src/app/api/cron/backfill/route.ts
+++ b/src/app/api/cron/backfill/route.ts
@@ -302,7 +302,7 @@ async function processStorylineCreated(
   };
   const { error: plotError } = await supabase
     .from("plots")
-    .upsert(plotRow, { onConflict: "tx_hash,log_index" });
+    .upsert(plotRow, { onConflict: "storyline_id,plot_index", ignoreDuplicates: true });
   if (plotError) {
     throw new Error(`Database error (genesis plot): ${plotError.message}`);
   }
@@ -360,7 +360,7 @@ async function processPlotChained(
 
   const { error: plotError } = await supabase
     .from("plots")
-    .upsert(row, { onConflict: "tx_hash,log_index" });
+    .upsert(row, { onConflict: "storyline_id,plot_index", ignoreDuplicates: true });
   if (plotError) {
     throw new Error(`Database error (plot): ${plotError.message}`);
   }
--- a/src/app/api/cron/backfill/route.ts
+++ b/src/app/api/cron/backfill/route.ts
@@ -360,7 +360,7 @@ async function processPlotChained(
 
   const { error: plotError } = await supabase
     .from("plots")
-    .upsert(row, { onConflict: "storyline_id,plot_index", ignoreDuplicates: true });
+    .upsert(row, { onConflict: "storyline_id,plot_index" });
   if (plotError) {
     throw new Error(`Database error (plot): ${plotError.message}`);
   }
--- a/src/app/api/index/plot/route.ts
+++ b/src/app/api/index/plot/route.ts
@@ -129,7 +129,7 @@ export async function POST(req: Request) {
 
   const { error: dbError } = await supabase.from("plots").upsert(
     row,
-    { onConflict: "storyline_id,plot_index", ignoreDuplicates: true }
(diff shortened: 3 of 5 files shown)


project7-interns (approved):
**T2b APPROVE**

Well-structured fix across migration + indexer routes:

- **Migration**: Correct order — dedupe (keep MIN id), add unique constraint, then reconcile. `IS DISTINCT FROM` in the reconcile UPDATE is a nice touch to skip no-op rows.
- **Upsert changes**: Switching conflict key from `tx_hash,log_index` to `storyline_id,plot_index` with `ignoreDuplicates: true` directly addresses the root cause (same plot indexed via different tx combos). Safe because plot data is immutable per index on-chain.
- **Comment** in reconcile.ts documents the constraint dependency.

No concerns.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The unique-constraint direction makes sense for fixing duplicate plot counts, but the current conflict handling introduces a metadata regression for genesis plots.

### Findings
- [high] Genesis plot titles can now be lost permanently. The storyline/backfill genesis insert paths still write the `plots` row without a `title`, but this PR changes those inserts to `onConflict: "storyline_id,plot_index", ignoreDuplicates: true`. That means if the genesis row is created first by the storyline path, the later plot-indexer write that does include `title` is discarded, leaving the row at the default empty string forever.
  - File: `src/app/api/index/storyline/route.ts:162`
  - File: `src/app/api/cron/backfill/route.ts:292`
  - Suggestion: Either include `title` in the genesis plot rows written by the storyline/backfill paths, or keep the new conflict key but allow merge updates instead of `ignoreDuplicates: true` so the later richer row can populate missing fields.

### Decision
Requesting changes because the current PR fixes the count issue by turning duplicate genesis writes into no-ops, but that also drops legitimate metadata from the surviving row.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new conflict-handling split fixes the future genesis-write race, but the migration still drops metadata from already duplicated genesis plots.

### Findings
- [high] The migration deletes duplicate `(storyline_id, plot_index)` rows by keeping `MIN(id)`, which will preserve the earlier bare genesis row and delete the later duplicate row that carries the genesis `title`. The follow-up upsert fix only helps future replays; it does not repair the duplicated rows that already exist and are being deduplicated here.
  - File: `supabase/migrations/00030_dedupe_plots_unique_constraint.sql:7`
  - File: `src/app/api/index/storyline/route.ts:162`
  - File: `src/app/api/index/plot/route.ts:117`
  - Suggestion: Before deleting duplicates, merge non-empty metadata from the duplicate set into the survivor, or choose the survivor from the richer row set instead of blindly keeping the lowest `id`.

### Decision
Requesting changes because the PR still turns an existing data-integrity bug into a persistent metadata-loss bug during migration for already duplicated genesis plots.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up migration change resolves the remaining data-loss issue by deduplicating `(storyline_id, plot_index)` rows in a way that prefers richer genesis rows before the unique constraint is added.

### Findings
- None.

### Decision
Approving because the PR now both prevents future duplicate plot rows and preserves the better existing row during migration, which addresses the plot-count bug without the metadata regression from the earlier revision.

PR #345, opened 2026-03-19
[#344] Update StoryFactory address and deployment block after redeploy

realproject7:
## Summary
- Replace old StoryFactory address `0x66087c...` with `0xc278F4099298118efA8dF30DF0F4876632571948` in all constants
- Update deployment block from `43559145` to `43561137`
- Updated in: `lib/contracts/constants.ts`, `packages/sdk/src/constants.ts`, `.env.example`

## Test plan
- [x] `npm run typecheck` passes
- [x] `npm run build` passes
- [x] No remaining references to old address `0x66087c...`

Fixes #344

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #787, opened 2026-04-03
Add USD-denominated price chart

realproject7:
## Summary
- Adds `reserve_usd_rate` and `rate_source` columns to `trade_history` table (migration 00031)
- Both trade indexers (cron + webhook) now fetch and store PLOT/USD rate at index time via `lib/reserve-usd-rate.ts`
- `PriceChart.tsx` gains a USD/PLOT segmented toggle (defaults to USD when data available)
- Approximate data points rendered with dashed lines + footnote disclosure (per T2a review)
- Backfill script (`scripts/backfill-usd-rates.ts`) with tiered strategy: exact historical reads via archive-capable RPCs, falling back to approximate current-rate extrapolation

## Design Decisions
- **PLOT/USD derivation:** `priceForNextMint(PLOT_TOKEN)` in HUNT × HUNT/USD via 1inch spot oracle
- **Backfill accuracy:** Exact where archive RPC supports historical `eth_call`; approximate otherwise. Rows tagged with `rate_source` (`live`, `backfill_exact`, `backfill_approx`) for future re-backfill
- **NULL handling:** If USD rate fetch fails, `reserve_usd_rate` = NULL, chart falls back to reserve-only display
- **No archive node dependency:** Works with public RPCs; some (mainnet.base.org, base.drpc.org) happen to support historical reads

## Approximate Data Disclosure
Per T2a review requirement: approximate USD data points are visually distinguished with dashed line segments and reduced opacity, plus a footnote when the chart contains any approximate data.

## Test Plan
- [ ] Run migration against Supabase
- [ ] Verify cron indexer stores `reserve_usd_rate` and `rate_source='live'` on new trades
- [ ] Verify webhook indexer stores USD rate on real-time trades
- [ ] Test PriceChart toggle between USD and PLOT modes
- [ ] Test chart with NULL `reserve_usd_rate` rows (should fall back to reserve-only)
- [ ] Run `npx tsx scripts/backfill-usd-rates.ts` to backfill existing trades
- [ ] Verify dashed line rendering for `backfill_approx` data points

Fixes #345

### DIFF
--- a/lib/contracts/abi.ts
+++ b/lib/contracts/abi.ts
@@ -132,6 +132,21 @@ export const priceForNextMintFunction = {
   outputs: [{ name: "", type: "uint128" }],
 } as const;
 
+/** 1inch Spot Price Aggregator: get exchange rate between two tokens. */
+export const spotPriceAbi = [
+  {
+    inputs: [
+      { name: "srcToken", type: "address" },
+      { name: "dstToken", type: "address" },
+      { name: "useWrappers", type: "bool" },
+    ],
+    name: "getRate",
+    outputs: [{ name: "weightedRate", type: "uint256" }],
+    stateMutability: "view",
+    type: "function",
+  },
+] as const;
+
 /** Full bond info for a token: creator, royalties, creation time, reserve. */
 export const tokenBondFunction = {
   type: "function",
--- a/lib/contracts/constants.ts
+++ b/lib/contracts/constants.ts
@@ -51,6 +51,9 @@ export const RESERVE_LABEL = "PLOT";
 // Supported Zap input tokens (Base)
 // ---------------------------------------------------------------------------
 
+/** 1inch Spot Price Aggregator on Base */
+export const ONEINCH_SPOT_PRICE_AGGREGATOR = "0x00000000000D6FFc74A8feb35aF5827bf57f6786" as const;
+
 /** USDC on Base */
 export const USDC = "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913" as const;
 
--- a/lib/reserve-usd-rate.ts
+++ b/lib/reserve-usd-rate.ts
@@ -9,25 +9,14 @@
 
 import { formatEther } from "viem";
 import { publicClient } from "./rpc";
-import { MCV2_BOND, PLOT_TOKEN, HUNT } from "./contracts/constants";
-import { priceForNextMintFunction } from "./contracts/abi";
-
-const USDC_ADDRESS = "0x833589fcd6edb6e08f4c7c32d4f71b54bda02913" as const;
-const ONEINCH_SPOT_PRICE_AGGREGATOR = "0x00000000000D6FFc74A8feb35aF5827bf57f6786" as const;
-
-const SPOT_PRICE_ABI = [
-  {
-    inputs: [
-      { name: "srcToken", type: "address" },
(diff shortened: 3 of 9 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The overall direction is sound, but the current implementation still writes materially incorrect USD history in a couple of paths. I would not ship this as-is because it can present stale or miscomputed USD prices as if they were historical truth.

### Findings
- [high] The approximate backfill path does not implement the strategy described in the PR. `getCurrentPlotUsd()` fetches a single current PLOT/USD rate and reuses that same value for every historical block, so rows marked `backfill_approx` ignore historical HUNT/USD entirely.
  - File: `scripts/backfill-usd-rates.ts:91`
  - Suggestion: fetch historical HUNT/USD for each trade date/block in the approximate path, or explicitly rename/reframe this source as a current-rate extrapolation instead of a historical approximation.

- [medium] The cron indexer fetches one current PLOT/USD rate per 200-block scan and stamps it onto every trade in that batch as `live`. If the cursor is catching up after downtime, those rows can be several minutes or more away from the actual trade time while still being labeled as exact live data.
  - File: `src/app/api/cron/trade-history/route.ts:90`
  - Suggestion: fetch/cache the reserve USD rate per unique block (or per event timestamp bucket), or mark batch-stamped rows differently so they are not indistinguishable from exact-at-trade-time data.

- [medium] In USD mode the chart filters out rows with `reserve_usd_rate = null`, then uses the last remaining USD point as `latest`. If the newest trades are missing a USD rate, the component will silently hide those rows and label an older point as the latest price.
  - File: `src/components/PriceChart.tsx:132`
  - Suggestion: do not collapse missing-rate rows out of the timeline when presenting a latest USD price. Either show a visible gap/stale state, or only default to USD when the newest displayed trade has a USD rate.

### Decision
Requesting changes because the current data pipeline can write or display inaccurate USD history without clearly communicating that loss of precision.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Two of my prior concerns look addressed, but one correctness issue remains in the cron path. The current catch-up detection still allows materially stale USD rates to be labeled as `live` for part of the final backlog batch.

### Findings
- [medium] `isCatchUp` is derived from `currentBlock - toBlock`, but the batch writes one shared `reserve_usd_rate` / `rate_source` across the whole `[fromBlock, toBlock]` window. In the final catch-up batch, `toBlock` can be within 200 blocks of head while `fromBlock` is still nearly 400 blocks behind, so older trades in that same batch are stamped as `live` even though they are not near-head.
  - File: `src/app/api/cron/trade-history/route.ts:94`
  - Suggestion: base the label on the oldest block in the batch (for example `currentBlock - fromBlock`), or compute/cache the reserve USD rate per event block / smaller bucket so that only truly near-head trades get `live`.

### Decision
Keeping this in request-changes until the cron labeling semantics match the precision of the data being written.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The remaining cron-labeling issue is fixed. Using `fromBlock` for the catch-up check is the conservative behavior I was asking for, and I do not have further correctness blockers on this PR.

### Findings
- None.

### Decision
Approved based on the current implementation and the issues addressed in the follow-up revisions.

PR #825, opened 2026-04-04
[#822] Replace dual-render with CSS grid-template-areas

### LINKED ISSUE #822: [Bug] Storyline header: stats grid rendered twice in DOM (hidden/block duplication)
## Problem
PR #821 uses `hidden sm:block` + `sm:hidden` to show stats in different positions on mobile vs desktop. This means the stats grid and CTA button are rendered **twice in the DOM** — one copy hidden via CSS. Components like `MarketCapBox` (API call) and `DeadlineCountdown` (setInterval timer) mount twice, doubling their side effects.

## Solution
Replace the dual-render approach with CSS grid and `grid-template-areas` so stats are rendered **once** and repositioned responsively.

```
<!-- Single render, repositioned via grid areas -->
<div class="grid grid-cols-[100px_1fr] sm:grid-cols-[160px_1fr] gap-x-4 sm:gap-x-6
            grid-rows-[auto_auto] sm:grid-rows-[1fr]
            [grid-template-areas:'cover_info'_'stats_stats'] 
            sm:[grid-template-areas:'cover_info'_'._stats']">
  
  <div style="grid-area: cover">Moleskine cover</div>
  <div style="grid-area: info">Title, rating, writer, genre</div>
  <div style="grid-area: stats">Stats grid + CTA</div>
</div>
```

Mobile: stats span full width below cover+info row (`stats stats`)
Desktop: stats align with info column only (`. stats` — dot = empty cell under cover)

This keeps the exact same visual result but with a single DOM instance of each component.

## Safety
- Pure CSS structural change — no logic, prop, or data changes
- Same visual output on both breakpoints
- MarketCapBox and DeadlineCountdown mount once instead of twice
- Reduces DOM size and eliminates duplicate API calls/timers

## Acceptance Criteria
- [ ] Stats grid and CTA rendered once in the DOM (no hidden duplicates)
- [ ] Mobile: stats full-width below cover+info (same as current)
- [ ] Desktop: stats in right column next to cover (same as current)
- [ ] Desktop spacing preserved (mt-6, gap-6)
- [ ] MarketCapBox and DeadlineCountdown only mount once
- [ ] No visual changes on any breakpoint

## Branch
`task/<issue>-header-single-render`

realproject7:
## Summary
- Replaced `hidden sm:block` + `sm:hidden` dual-render pattern with CSS `grid-template-areas`
- Stats grid and CTA button now rendered **once** in the DOM instead of twice
- Mobile: stats span full width below cover+info row (`'cover info' / 'stats stats'`)
- Desktop: stats align with info column only (`'cover info' / '. stats'`)
- MarketCapBox and DeadlineCountdown now mount once (no duplicate API calls/timers)
- No visual changes on any breakpoint

## Test plan
- [ ] Desktop: stats grid + CTA next to cover in right column
- [ ] Mobile: stats grid full-width below cover+info row
- [ ] Mobile: CTA button full-width
- [ ] Desktop: CTA button inline/auto-width
- [ ] Inspect DOM: stats grid and CTA appear only once
- [ ] Verify MarketCapBox network calls fire once (not twice)

Fixes #822

### DIFF
--- a/src/app/story/[storylineId]/page.tsx
+++ b/src/app/story/[storylineId]/page.tsx
@@ -315,82 +315,76 @@ function StoryHeader({
   );
 
   return (
-    <header className="pb-6">
-      <div className="flex flex-row items-start gap-4 sm:gap-6">
-        {/* Moleskine book cover */}
-        <div className="shrink-0 w-[100px] sm:w-[160px]">
+    <header
+      className="pb-6 grid gap-x-4 sm:gap-x-6 grid-cols-[100px_1fr] sm:grid-cols-[160px_1fr] [grid-template-areas:'cover_info'_'stats_stats'] sm:[grid-template-areas:'cover_info'_'._stats']"
+    >
+      {/* Moleskine book cover */}
+      <div className="[grid-area:cover]">
+        <div
+          className="relative flex flex-col overflow-hidden border border-[var(--border)]"
+          style={{
+            aspectRatio: "2/3",
+            borderRadius: "5px 12px 12px 5px",
+            backgroundColor: "#F5EFE6",
+            boxShadow: "2px 4px 8px rgba(44, 24, 16, 0.08)",
+          }}
+        >
           <div
-            className="relative flex flex-col overflow-hidden border border-[var(--border)]"
-            style={{
-              aspectRatio: "2/3",
-              borderRadius: "5px 12px 12px 5px",
-              backgroundColor: "#F5EFE6",
-              boxShadow: "2px 4px 8px rgba(44, 24, 16, 0.08)",
-            }}
-          >
-            <div
-              className="pointer-events-none absolute inset-y-[-1px] right-[16px] z-20 w-[5px] rounded-[2px]"
-              style={{ background: "rgba(139, 69, 19, 0.15)" }}
-            />
-            <div className="relative z-10 px-2.5 pt-2.5">
-              <span className="rounded-sm bg-[var(--accent)]/10 px-1.5 py-0.5 text-[8px] font-semibold uppercase tracking-widest text-[var(--accent)]">
-                {storyline.genre || "Uncategorized"}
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
LGTM. Grid-template-areas approach correctly eliminates the dual-render pattern. Stats and CTA mount once, grid columns preserve the 100px/160px cover sizing, and the responsive `sm:` reset on CTA width (`w-auto`) matches the previous default. No visual or logic regressions.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The single-render grid approach is the right fix for issue #822, but this version introduces a layout regression when `priceInfo` is missing.

### Findings
- [medium] The new `stats` grid item is always rendered with `mt-4 sm:mt-6`, even when both `statsGrid` and `AddPlotButton` render nothing. On untokenized stories, that leaves an empty spacer below the header for all viewers, and when only the CTA is present it adds extra top spacing versus the old layout.
  - File: `src/app/story/[storylineId]/page.tsx:384`
  - Suggestion: Keep the grid-area wrapper marginless and apply the responsive top margin only to `statsGrid`, so the CTA-only / empty states preserve the previous spacing.

### Decision
Requesting changes because the issue requires no visual regressions, and this change alters the header spacing for stories without price data.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up commit fixes the header spacing regression from the earlier review while preserving the single-render grid-template-areas solution for issue #822.

### Findings
- None. The `mt-4 sm:mt-6` spacing now applies only when `statsGrid` exists, so untokenized stories no longer get empty spacer height and CTA-only states keep their previous spacing.

### Decision
Approving because the PR now meets the acceptance criteria: stats and CTA render once in the DOM, side effects mount once, and the prior visual regression is resolved.

PR #834, opened 2026-04-04
[#824] Fix Writer tab badge to use time-based deadline expiry

### LINKED ISSUE #824: [Bug] Writer tab: 'active' badge shows on expired storylines
## Problem
In the Writer tab (profile page), the status badge shows "active" even when the deadline text shows "expired". Same root cause as #804/#808 — the badge checks `sunset` flag (never set by app code) instead of the actual time-based deadline expiry.

Screenshot shows: "Deadline: expired · active" — contradictory.

## Solution
Use the same time-based check from `DeadlineCountdown.tsx`:
- If `has_deadline && last_plot_time + 168h < now` → show "expired" badge (red/error style)
- If `sunset` → show "complete" badge
- Otherwise → show "active" badge (green)

Check the profile page Writer tab storyline card rendering in `src/app/profile/[address]/page.tsx` for the badge logic.

## Acceptance Criteria
- [ ] Badge shows "expired" when deadline has passed (time-based check)
- [ ] Badge shows "complete" when `sunset=true`
- [ ] Badge shows "active" only when storyline is truly active
- [ ] Uses `has_deadline` guard (storylines without deadlines are always "active")

## Branch
`task/<issue>-writer-active-badge`

realproject7:
## Summary
- Badge now checks `has_deadline && last_plot_time + 168h < now` instead of only `sunset`
- Shows "expired" (red) when deadline has passed, "complete" when sunset, "active" (green) otherwise
- Imports `DEADLINE_MS` from `DeadlineCountdown.tsx` for consistent deadline logic
- Fixes contradictory "Deadline: expired · active" display

## Test plan
- [ ] Storyline with expired deadline shows "expired" badge (red)
- [ ] Storyline with `sunset=true` shows "complete" badge
- [ ] Active storyline (within deadline) shows "active" badge (green)
- [ ] Storyline without `has_deadline` shows "active" badge

Fixes #824

### DIFF
--- a/src/app/profile/[address]/page.tsx
+++ b/src/app/profile/[address]/page.tsx
@@ -868,9 +868,18 @@ function StoryRow({
     enabled: !!storyline.token_address,
   });
 
-  const [now] = useState(() => Date.now());
-  const isExpired = !storyline.sunset && storyline.has_deadline && !!storyline.last_plot_time &&
-    now > new Date(storyline.last_plot_time).getTime() + DEADLINE_MS;
+  const checkExpired = useCallback(
+    () => !storyline.sunset && storyline.has_deadline && !!storyline.last_plot_time &&
+      Date.now() > new Date(storyline.last_plot_time).getTime() + DEADLINE_MS,
+    [storyline.sunset, storyline.has_deadline, storyline.last_plot_time],
+  );
+  const [isExpired, setIsExpired] = useState(checkExpired);
+  useEffect(() => {
+    // eslint-disable-next-line react-hooks/set-state-in-effect -- initial sync needed for SSR hydration safety
+    setIsExpired(checkExpired());
+    const interval = setInterval(() => setIsExpired(checkExpired()), 60_000);
+    return () => clearInterval(interval);
+  }, [checkExpired]);
 
   return (
     <>
--- a/src/app/profile/[address]/page.tsx
+++ b/src/app/profile/[address]/page.tsx
@@ -19,7 +19,7 @@ import { usePlotUsdPrice } from "../../../hooks/usePlotUsdPrice";
 import { formatUsdValue } from "../../../../lib/usd-price";
 import { DisconnectButton } from "../../../components/ConnectWallet";
 import { GENRES, LANGUAGES } from "../../../../lib/genres";
-import { DeadlineCountdown } from "../../../components/DeadlineCountdown";
+import { DeadlineCountdown, DEADLINE_MS } from "../../../components/DeadlineCountdown";
 import { ClaimRoyalties } from "../../../components/ClaimRoyalties";
 import { WriterTradingStats } from "../../../components/WriterTradingStats";
 import { DropdownSelect } from "../../../components/DropdownSelect";
@@ -952,6 +952,8 @@ function StoryRow({
           )}
           {storyline.sunset ? (
             <span className="border-border text-muted rounded border px-1.5 py-0.5 text-[10px]">complete</span>
+          ) : storyline.has_deadline && storyline.last_plot_time && Date.now() > new Date(storyline.last_plot_time).getTime() + DEADLINE_MS ? (
+            <span className="border border-red-700/30 text-red-700 rounded px-1.5 py-0.5 text-[10px]">expired</span>
           ) : (
             <span className="border border-green-700/30 text-green-700 rounded px-1.5 py-0.5 text-[10px]">active</span>
           )}
--- a/src/app/profile/[address]/page.tsx
+++ b/src/app/profile/[address]/page.tsx
@@ -868,6 +868,9 @@ function StoryRow({
     enabled: !!storyline.token_address,
   });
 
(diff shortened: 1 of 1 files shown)


project7-interns (approved):
LGTM. Badge priority order is correct: sunset→complete, then time-based deadline check→expired, else→active. Reuses the shared `DEADLINE_MS` constant from `DeadlineCountdown.tsx` (168h) and matches the same pattern already used in `create/page.tsx`. The `has_deadline` guard correctly keeps non-deadline storylines as 'active'.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The Writer tab badge now uses the intended time-based deadline expiry check instead of relying only on `sunset`, which resolves the contradictory expired-vs-active display from issue #824.

### Findings
- None. The change is narrowly scoped to `src/app/profile/[address]/page.tsx`, reuses `DEADLINE_MS` from `DeadlineCountdown.tsx`, and preserves the existing `sunset` -> `complete` path.

### Decision
Approving because the PR matches the acceptance criteria: deadline-based stories can now show `expired`, `sunset` still maps to `complete`, and non-deadline stories remain `active`.

project7-interns (approved):
Re-approved after 6df20e7. `Date.now()` correctly hoisted to `isExpired` variable before return, resolving the react-hooks/purity lint. Logic unchanged — same priority order and guards.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The follow-up commit does not actually resolve the lint blocker on PR #834. `Date.now()` was moved out of JSX, but it is still being called during `StoryRow` render.

### Findings
- [medium] `isExpired` still calls `Date.now()` inside the component body, which is still render-time impurity and still fails `react-hooks/purity`. The current GitHub Actions run already shows `lint-and-typecheck` failing on this change.
  - File: `src/app/profile/[address]/page.tsx:871`
  - Suggestion: Avoid calling `Date.now()` during render at all. Derive the reference time outside render (for example via state/effect or another existing time source) and then compare that value against `last_plot_time + DEADLINE_MS`.

### Decision
Requesting changes because the PR remains unmergeable: the intended badge logic is fine, but the latest implementation still violates the purity lint rule.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new `useState(() => Date.now())` fix avoids render-time impurity, but it freezes the reference time at mount and makes the Writer-tab badge drift out of sync with the live deadline countdown.

### Findings
- [medium] `now` is captured once with `useState(() => Date.now())` and never updates, so a storyline that is still active when the page mounts will keep its green `active` badge even after the deadline passes. At that point `DeadlineCountdown` will flip to `expired` while the badge stays `active`, recreating the contradictory state this issue is supposed to fix.
  - File: `src/app/profile/[address]/page.tsx:871`
  - Suggestion: Use a time source that updates after mount instead of a frozen timestamp, so the badge can transition from `active` to `expired` while the page is open without violating the purity lint rule.

### Decision
Requesting changes because the current implementation fixes the lint symptom but reintroduces the underlying expired-vs-active mismatch once the deadline passes during an open session.

project7-interns (approved):
Re-approved after 02a8f93. Live-updating `useCallback` + `useState` + `useEffect`/`setInterval` (60s) pattern matches `DeadlineCountdown.tsx`. Badge will correctly transition from active→expired in real time. The `eslint-disable` for the initial-sync `setIsExpired` follows the existing codebase convention.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new interval-based fix removes the frozen timestamp problem, but it still leaves a visible contradiction window because the badge updates much less frequently than `DeadlineCountdown`.

### Findings
- [medium] The Writer-tab badge only recomputes expiry once every `60_000ms`, while `DeadlineCountdown` updates every second. For up to 59 seconds after the deadline passes, the row can still show `Deadline: expired · active`, which is the exact contradiction issue #824 is meant to eliminate.
  - File: `src/app/profile/[address]/page.tsx:880`
  - Suggestion: Keep the badge on the same cadence as `DeadlineCountdown` or derive both from a shared time source so they flip to `expired` together.

### Decision
Requesting changes because the current implementation fixes the stale mount-time timestamp but still allows the expired-vs-active mismatch during an open session.

project7-interns (approved):
Re-approved after 5d06c59. Interval now 1s matching DeadlineCountdown cadence — badge and countdown flip together with no contradiction window.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The latest commit aligns the Writer-tab badge update cadence with `DeadlineCountdown`, so the badge can transition from `active` to `expired` in sync without reintroducing the render-purity lint failure.

### Findings
- None. The `useCallback` + `useState` + `useEffect` interval approach now ticks every `1_000ms`, matching `DeadlineCountdown` and removing the contradiction window from the prior version.

### Decision
Approving because the PR now meets the acceptance criteria: expired deadlines show `expired`, `sunset` still maps to `complete`, non-deadline stories stay `active`, and the badge/countdown stay in sync during an open session.

PR #854, opened 2026-04-12
[#835] Fix SSR hydration mismatch in Writer tab expired badge

### LINKED ISSUE #835: [Bug] Writer tab: useState(checkExpired) causes SSR hydration mismatch
## Problem
PR #834 uses `useState(checkExpired)` as a lazy initializer, which calls `Date.now()` during the initial render. Server and client timestamps differ, causing a potential React hydration mismatch warning for storylines near the expiry boundary.

Also: the 1-second `setInterval` polling is wasteful for a badge that only transitions once (active → expired). A `setTimeout` targeting the exact expiry time would be more efficient.

## Solution
1. Change `useState(checkExpired)` to `useState(false)` — let the existing `useEffect` set the correct value on mount
2. Replace `setInterval(1000)` with `setTimeout` calculated from `expiryTime - Date.now()`

## Acceptance Criteria
- [ ] No SSR hydration mismatch warnings
- [ ] Badge still correctly shows expired/active/complete
- [ ] No 1-second polling — use targeted setTimeout instead
- [ ] Cleanup timeout on unmount

## Branch
`task/<issue>-badge-hydration-fix`

realproject7:
## Summary
- Changed `useState(checkExpired)` to `useState(false)` — avoids calling `Date.now()` during server render which caused hydration mismatch near expiry boundary
- Replaced 1-second `setInterval` polling with a single `setTimeout` targeting the exact expiry time — fires once when the storyline actually expires
- Proper cleanup on unmount
- Patch version bump 0.1.19 → 0.1.20

Fixes #835

## Test plan
- [ ] No React hydration mismatch warnings in console
- [ ] Badge shows "active" for non-expired storylines, transitions to "expired" at the right time
- [ ] Badge shows "complete" for sunset storylines
- [ ] Badge shows "expired" immediately for already-expired storylines on page load

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "plotlink",
-  "version": "0.1.19",
+  "version": "0.1.20",
   "private": true,
   "workspaces": [
     "packages/*"
--- a/src/app/profile/[address]/page.tsx
+++ b/src/app/profile/[address]/page.tsx
@@ -886,7 +886,11 @@ function StoryRow({
 
   const [isExpired, setIsExpired] = useState(false);
   useEffect(() => {
-    if (storyline.sunset || !storyline.has_deadline || !storyline.last_plot_time) return;
+    if (storyline.sunset || !storyline.has_deadline || !storyline.last_plot_time) {
+      // eslint-disable-next-line react-hooks/set-state-in-effect -- reset when props change (e.g. deadline extension)
+      setIsExpired(false);
+      return;
+    }
     const expiryTime = new Date(storyline.last_plot_time).getTime() + DEADLINE_MS;
     const remaining = expiryTime - Date.now();
     if (remaining <= 0) {
@@ -894,6 +898,8 @@ function StoryRow({
       setIsExpired(true);
       return;
     }
+    // eslint-disable-next-line react-hooks/set-state-in-effect -- reset in case props changed from expired to active
+    setIsExpired(false);
     const timeout = setTimeout(() => setIsExpired(true), remaining);
     return () => clearTimeout(timeout);
   }, [storyline.sunset, storyline.has_deadline, storyline.last_plot_time]);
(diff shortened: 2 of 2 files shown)


realproject7 (commented):
**re2 APPROVE** — Clean fix. `useState(false)` eliminates the SSR/client divergence, and the targeted `setTimeout` is a big improvement over 1-second polling. Edge cases (already expired, sunset, no deadline) handled correctly. Cleanup on unmount is proper.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The SSR fix is on the right track, but the new timeout-based effect introduces stale state: `isExpired` is never reset back to `false` when the underlying storyline becomes non-expired.

### Findings
- [medium] `isExpired` can remain stuck `true` after storyline data changes
  - File: `src/app/profile/[address]/page.tsx:887`
  - Suggestion: Reset `isExpired` inside the effect before the early returns / timeout setup, or otherwise derive the current non-expired state when `storyline.sunset`, `storyline.has_deadline`, or `storyline.last_plot_time` changes. Right now a row that was previously expired can keep rendering the expired badge after a deadline extension or prop change.

### Decision
Requesting changes because the hydration fix should not introduce stale badge state on rerenders or data refreshes.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up addresses the stale state regression: `isExpired` is now reset when storyline props change, while still avoiding the SSR hydration mismatch and interval polling.

### Findings
- [info] GitHub checks are still pending at review time (`e2e`, `lint-and-typecheck`, `visual-regression`)
  - File: `PR #854`
  - Suggestion: Let CI finish before merge.

### Decision
Approved because the revised effect now preserves correct badge behavior across prop changes and matches issue #835 without reintroducing the original hydration problem.

PR #868, opened 2026-04-12
[#867] Distinguish direct agents from linked OWS writer owners

### LINKED ISSUE #867: [Follow-up] Distinguish direct agents from linked OWS writer owners
## Scope
Fix the owner-vs-agent profile classification introduced in #862 so direct agents are not misclassified as human owner profiles.

Observed risk:
- `isAgentOwner` currently treats an address as a human owner when `agentMeta.owner === address` and `agent_wallet` is null or different.
- That is correct for linked OWS writers, but it can also match direct agents whose registration persists `agent_owner = owner wallet` before any separate `agent_wallet` exists.

## Acceptance Criteria
- Direct agents continue to render as AI agent profiles, not human owner profiles
- Linked OWS writer owners continue to render as human profiles with a linked AI writer card
- Owner/agent classification logic explicitly distinguishes direct-agent registration from linked OWS writer ownership
- Profile and writer identity surfaces remain consistent with the intended linked AI writer behavior from #862

realproject7:
## Summary
- Adds `agent_type` column (`'direct'` / `'ows-writer'`) to the `users` table to explicitly classify agent registrations
- Updates `isAgentOwner` logic in `getFullUserProfile()` and `getAgentOwnerProfile()` to use `agent_type` instead of the overly broad wallet-null heuristic
- Direct agents now correctly render as AI agent profiles; OWS-linked owners continue to render as human profiles with linked AI writer cards

## Changes
- **Migration**: `00032_users_agent_type.sql` — adds `agent_type TEXT` column
- **Supabase types**: Row/Insert/Update types updated with `agent_type`
- **API** (`agent-register/route.ts`): accepts and persists `agentType`
- **Registration** (`AgentRegister.tsx`): OWS flow sends `agentType: "ows-writer"`, direct flow sends `agentType: "direct"`
- **Classification** (`lib/actions.ts`): `isAgentOwner` now checks `agent_type === "ows-writer"` instead of wallet-null heuristic
- **Version**: patch bump 0.1.23 → 0.1.24

## Test plan
- [ ] Register a direct agent → profile page should show "Agent Identity" card (not "Linked AI Writer")
- [ ] Register an OWS-linked writer → owner profile should show human identity with "Linked AI Writer" card
- [ ] Existing agents without `agent_type` (null) should default to agent display (not misclassified as owner)
- [ ] Agent management page (`/agents`) continues to work for both registration types

Fixes #867

### DIFF
--- a/lib/actions.ts
+++ b/lib/actions.ts
@@ -111,10 +111,12 @@ export async function getFullUserProfile(
 
   // Detect if this address is the agent OWNER (not the agent itself).
   // When true, the profile should display human identity, not agent identity.
+  // Direct agents (agent_type='direct') are the agent themselves — isAgentOwner stays false.
+  // OWS-linked writers (agent_type='ows-writer') have a separate human owner — isAgentOwner is true.
   const normalized = address.toLowerCase();
   const isAgentOwner = agentMeta !== null
     && agentMeta.owner?.toLowerCase() === normalized
-    && (dbUser?.agent_wallet == null || dbUser.agent_wallet.toLowerCase() !== normalized);
+    && dbUser?.agent_type === "ows-writer";
 
   return { dbUser, fcProfile, agentMeta, isAgentOwner };
 }
@@ -289,12 +291,12 @@ export async function getAgentOwnerProfile(
   const agentUser = await getAgentUserFromDB(writerAddress);
   if (!agentUser?.agent_id) return null;
 
-  // If the queried address IS the agent owner (not the agent wallet),
+  // If the queried address is an OWS-linked owner (not the agent itself),
   // this address belongs to the human owner — not an agent.
+  // Direct agents (agent_type='direct') are the agent themselves and should not be excluded.
   const normalized = writerAddress.toLowerCase();
   const isOwnerAddress = agentUser.agent_owner?.toLowerCase() === normalized;
-  const isAgentWallet = agentUser.agent_wallet?.toLowerCase() === normalized;
-  if (isOwnerAddress && !isAgentWallet) return null;
+  if (isOwnerAddress && agentUser.agent_type === "ows-writer") return null;
 
   const ownerProfile = agentUser.agent_owner
     ? await getFarcasterProfile(agentUser.agent_owner)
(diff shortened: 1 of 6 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new `agent_type` discriminator fixes the direct-agent misclassification for newly registered records, but it introduces a regression for existing linked OWS writer rows because the migration does not backfill `agent_type`.

### Findings
- [high] Existing linked OWS writer records remain `agent_type = NULL`, so the new owner-classification checks stop recognizing them as linked-owner profiles.
  - File: `lib/actions.ts:118`
  - Suggestion: Either backfill existing OWS-linked rows in the migration, or preserve a compatibility fallback for pre-existing rows until the data is migrated.
- [high] The schema change adds a nullable column only, so deployed data keeps its old shape and the runtime logic change takes effect immediately against unclassified rows.
  - File: `supabase/migrations/00032_users_agent_type.sql:4`
  - Suggestion: Add a data migration that marks existing linked OWS writer rows as `ows-writer` where appropriate, or make the runtime logic resilient to null legacy rows.

### Decision
Requesting changes because this PR meets the new-registration path but does not preserve the existing linked OWS writer behavior required by the issue acceptance criteria.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The legacy fallback fixes the regression I flagged on the previous revision, but the new discriminator is still not validated at the API boundary.

### Findings
- [medium] `agent-register` persists arbitrary `agentType` values, so callers can store strings other than `"direct"` or `"ows-writer"` and break the classification logic the PR is introducing.
  - File: `src/app/api/user/agent-register/route.ts:11`
  - Suggestion: Reject invalid `agentType` values with a 400, or normalize through a narrow allowlist before writing to the database.

### Decision
Requesting changes because the classification fix depends on `agent_type` containing only the two supported values, and the current API does not enforce that invariant.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The PR now preserves legacy linked OWS writer behavior and constrains the new `agent_type` discriminator to the supported values for new writes.

### Findings
- No further code findings.

### Decision
Approving. The remaining GitHub checks were still pending at review time, but the code issues raised in prior review rounds are resolved.

PR #896, opened 2026-04-20
[#876] Fix Market Cap blank box during price loading

### LINKED ISSUE #876: Market Cap stat box renders blank while PLOT/USD price is loading
## Problem

On the story page (`/story/[storylineId]`), the first stat box (Market Cap) renders as an empty bordered box while the PLOT/USD price is loading. After a few seconds it eventually appears (e.g. "$68.84 +0.0%"), meaning all 3 price sources in the fallback chain are slow or failing before one finally succeeds.

## Root Cause

Two issues:

### 1. Price API fallback chain is slow/unreliable

`getPlotUsdPrice()` in `lib/usd-price.ts` tries 3 sources **sequentially** with 3s timeouts each:

1. **Mint Club SDK** — dynamic import + RPC call. Likely fails or times out (3s wasted)
2. **GeckoTerminal** — HTTP fetch. May not list PLOT or may be rate-limited (3s wasted)
3. **CoinGecko** — HTTP fetch. May not list PLOT without API key (3s wasted)

Worst case: 9 seconds of sequential timeouts before returning `null`. Even if one source eventually works, the user sees a blank box for several seconds.

**Investigation needed:**
- Which source(s) actually succeed for the PLOT token?
- Which ones consistently fail? (check server logs for `[USD Price] source=X result=miss`)
- Are we hitting rate limits on GeckoTerminal/CoinGecko?
- Is the Mint Club SDK import failing on the server side?
- Is the 2-minute in-memory cache (`CACHE_TTL`) working, or is it being invalidated between requests (e.g. serverless cold starts)?

### 2. Loading state shows blank box

`MarketCapBox` (`src/components/MarketCapBox.tsx:30`) returns `null` while `plotUsd` is loading:

```ts
if (!plotUsd) return null;
```

The parent always renders the box container, so `null` content = empty bordered box.

### Part A: Diagnose & fix the price source reliability

1. **Add structured logging** to `fetchPlotUsdPrice()` — log which source succeeded, how long each attempt took, and why failures occurred (timeout vs HTTP error vs missing data)
2. **Identify which sources actually work** for PLOT token and remove/deprioritize dead sources
3. **Consider `Promise.any()`** instead of sequential fallback — try all sources in parallel, use whichever responds first
4. **Extend cache TTL or add stale-while-revalidate** — return stale cached price immediately while refreshing in background, especially important for serverless where in-memory cache is lost on cold starts
5. **Consider DB-backed cache** as a last-resort fallback (store last known good price in Supabase) so there's always a value to show

### Part B: Loading state fallback

Show "—" with label while loading, so the box is never blank:

```tsx
if (!plotUsd) {
  return (
    <>
      <div className="text-foreground text-sm font-bold">—</div>
      <div className="text-muted text-[9px]">Market Cap</div>
    </>
  );
}
```

## Files

- `lib/usd-price.ts` — price fetching logic, cache, fallback chain
- `src/components/MarketCapBox.tsx` — loading state UI
- `src/hooks/usePlotUsdPrice.ts` — client-side hook
- `src/app/api/tokens/plot-price/route.ts` — API endpoint

## Acceptance Criteria

- [ ] Logs identify which price sources work/fail and why
- [ ] Dead or consistently-failing sources are removed or deprioritized
- [ ] Price loads fast enough that users rarely see the fallback state
- [ ] Market Cap box never renders blank — shows "—" while loading
- [ ] Cached price survives serverless cold starts (DB fallback or extended cache)

realproject7:
## Summary
- Switched PLOT/USD price sources from sequential fallback (up to 9s worst case) to `Promise.any()` parallel fetch — uses whichever source responds first
- Added DB fallback (`trade_history.reserve_usd_rate`) so cached price survives serverless cold starts
- Added structured logging with per-source timing (`elapsed=Xms`) for diagnosing which sources work/fail
- MarketCapBox now shows "—" placeholder while loading instead of rendering a blank box

## Files Changed
- `lib/usd-price.ts` — parallel fetch, DB fallback, structured logging
- `src/components/MarketCapBox.tsx` — loading state placeholder
- `package.json` — patch bump 0.1.24 → 0.1.25

## Test plan
- [ ] Verify Market Cap box shows "—" on initial page load before price arrives
- [ ] Verify Market Cap value populates correctly after price loads
- [ ] Check server logs for `[USD Price] source=X result=hit/miss elapsed=Xms` entries
- [ ] Confirm price loads significantly faster than before (parallel vs sequential)
- [ ] Cold-start scenario: verify DB fallback kicks in when in-memory cache is empty and external sources are slow

Fixes #876

### DIFF
--- a/lib/usd-price.ts
+++ b/lib/usd-price.ts
@@ -80,19 +80,28 @@ async function fetchPlotUsdPrice(): Promise<number | null> {
   return null;
 }
 
-/** Mint Club SDK — on-chain RPC call */
+/** Mint Club SDK — on-chain RPC call (with 3s timeout to match other sources) */
 async function fetchFromMintClub(): Promise<number> {
   const start = Date.now();
   try {
-    const { mintclub } = await import(/* webpackIgnore: true */ "mint.club-v2-sdk" as string) as { mintclub: { network: (n: string) => { token: (a: `0x${string}`) => { getUsdRate: () => Promise<{ usdRate: number }> } } } };
-    const token = mintclub.network("base").token(PLOT_TOKEN);
-    const { usdRate } = await token.getUsdRate();
-    if (usdRate && usdRate > 0) {
+    const result = await Promise.race([
+      (async () => {
+        const { mintclub } = await import(/* webpackIgnore: true */ "mint.club-v2-sdk" as string) as { mintclub: { network: (n: string) => { token: (a: `0x${string}`) => { getUsdRate: () => Promise<{ usdRate: number }> } } } };
+        const token = mintclub.network("base").token(PLOT_TOKEN);
+        return token.getUsdRate();
+      })(),
+      new Promise<never>((_, reject) =>
+        setTimeout(() => reject(new Error("timeout")), 3000),
+      ),
+    ]);
+    if (result.usdRate && result.usdRate > 0) {
       console.info(`[USD Price] source=mint_club result=hit elapsed=${Date.now() - start}ms`);
-      return usdRate;
+      return result.usdRate;
     }
-  } catch {
-    console.info(`[USD Price] source=mint_club result=miss elapsed=${Date.now() - start}ms`);
+    throw new Error("invalid rate");
+  } catch (e) {
+    const reason = e instanceof Error ? e.message : "unknown";
+    console.info(`[USD Price] source=mint_club result=miss reason=${reason} elapsed=${Date.now() - start}ms`);
   }
   throw new Error("mint_club failed");
 }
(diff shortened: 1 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The loading placeholder fix in `MarketCapBox` is correct, but the new server-side price fetch path still has a blocker. As written, the DB fallback is not guaranteed to run during the exact cold-start / slow-source scenario this PR is trying to fix.

### Findings
- [high] `Promise.any()` can hang forever because `fetchFromMintClub()` has no timeout, so `fetchPlotUsdPrice()` never reaches the DB fallback when Mint Club stalls and the HTTP sources reject.
  - File: `lib/usd-price.ts:60`
  - Suggestion: wrap the Mint Club SDK path in the same bounded timeout policy as the HTTP fetches, or race each source against a timeout before passing them into `Promise.any()`.
- [medium] The issue asks for logs that identify why each source failed, but the new catch blocks only emit `result=miss` with elapsed time and discard the actual failure reason/status.
  - File: `lib/usd-price.ts:94`
  - Suggestion: log timeout vs HTTP status vs missing data so production logs can distinguish dead sources from transient failures.

### Decision
Requesting changes because the main reliability fix is incomplete: if Mint Club hangs, users can still get stuck without a price and without the advertised DB fallback.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The rework fixes the blocker from my previous review. `fetchFromMintClub()` is now bounded by a 3s timeout, so `Promise.any()` can fail fast and reach the DB fallback, and the source-level logging now captures concrete failure reasons.

### Findings
- No blocking findings in the updated patch.

### Decision
Approving from a code-review standpoint. The timeout and logging gaps I raised are addressed in `lib/usd-price.ts`, and the `MarketCapBox` placeholder change remains correct.

PR #909, opened 2026-04-21
[#890] Add daily price snapshot cron

### LINKED ISSUE #890: [Airdrop P4] Daily price snapshot cron
Parent: #877
Depends on: #878

## Scope

Daily cron job that records PLOT USD price, supply, and mcap for TWAP calculation.

## Logic

1. Fetch PLOT USD price via `getPlotUsdPrice()` (existing in `lib/usd-price.ts`)
2. Fetch circulating supply from MCV2 Bond or Mint Club SDK
3. Compute mcap = price × supply
4. Insert into `pl_daily_prices` (date-unique — skip if today already has an entry)

**Error handling:** If `getPlotUsdPrice()` returns `null` (all 3 sources failed), skip the day's snapshot — do NOT insert a zero or stale value. The TWAP calculation at campaign end uses `AVG()` over available days, so a missing day is better than a wrong value. Log the failure for monitoring.

## Endpoint

`POST /api/cron/airdrop-price`

**Schedule:** Once/day at midnight UTC (`0 0 * * *`)
**Auth:** `CRON_SECRET` header

## TWAP Usage

At campaign end, the final milestone is determined by the 7-day TWAP:
`AVG(mcap_usd) FROM pl_daily_prices WHERE recorded_at >= (end_date - 7 days)`

## Files

- `src/app/api/cron/airdrop-price/route.ts`

## Acceptance Criteria

- [ ] Records daily PLOT price + supply + mcap
- [ ] Idempotent — one entry per day
- [ ] Uses existing `getPlotUsdPrice()` (no new price source)
- [ ] Skips snapshot if price fetch fails (no zero/stale values)
- [ ] Logs failure when price fetch returns null
- [ ] Registered in `vercel.json` crons (see #891)

realproject7:
Fixes #890

## Summary
- **POST /api/cron/airdrop-price**: daily cron that records PLOT USD price, circulating supply, and mcap into `pl_daily_prices`
- Uses existing `getPlotUsdPrice()` for price and on-chain `totalSupply` via viem for supply
- Idempotent: checks for existing entry before insert (one per day)
- Skips snapshot if price or supply fetch fails (no zero/stale values) — logs failure
- Auth via `CRON_SECRET` Bearer token (matches existing cron pattern)
- Registered in `vercel.json` at `0 0 * * *` (midnight UTC)

## Test plan
- [ ] Verify cron rejects unauthorized requests (401)
- [ ] Verify idempotency — second call same day returns skipped
- [ ] Verify price null → skip with log, no DB insert
- [ ] Verify supply fetch failure → skip with log, no DB insert
- [ ] Verify successful insert with correct price, supply, mcap values
- [ ] Verify vercel.json cron schedule is correct

### DIFF
--- a/src/app/api/cron/airdrop-price/route.ts
+++ b/src/app/api/cron/airdrop-price/route.ts
@@ -1,0 +1,86 @@
+/**
+ * Daily price snapshot cron (#890)
+ *
+ * Records PLOT USD price, circulating supply, and mcap for TWAP calculation.
+ * Schedule: once/day at midnight UTC (0 0 * * *)
+ */
+
+import { NextResponse } from "next/server";
+import { formatUnits } from "viem";
+import { createServerClient } from "../../../../../lib/supabase";
+import { getPlotUsdPrice } from "../../../../../lib/usd-price";
+import { publicClient } from "../../../../../lib/rpc";
+import { PLOT_TOKEN } from "../../../../../lib/contracts/constants";
+import { erc20Abi } from "../../../../../lib/price";
+
+function verifyCron(req: Request): boolean {
+  const secret = process.env.CRON_SECRET;
+  if (!secret) {
+    return process.env.NODE_ENV !== "production";
+  }
+  const authHeader = req.headers.get("authorization");
+  return authHeader === `Bearer ${secret}`;
+}
+
+export async function POST(req: Request) {
+  if (!verifyCron(req)) {
+    return NextResponse.json({ error: "Unauthorized" }, { status: 401 });
+  }
+
+  const supabase = createServerClient();
+  if (!supabase) {
+    return NextResponse.json({ error: "Supabase not configured" }, { status: 500 });
+  }
+
+  const todayUtc = new Date().toISOString().slice(0, 10);
+
+  // Idempotency: skip if today already has an entry
+  const { data: existing } = await supabase
+    .from("pl_daily_prices")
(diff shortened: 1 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The snapshot logic is mostly aligned with issue #890, but the Vercel cron registration will not actually invoke this route as implemented.

### Findings
- [high] The route only exports `POST`, but `vercel.json` cron jobs call the configured path with `GET`. That means the scheduled job at `/api/cron/airdrop-price` will hit a method mismatch instead of recording a daily snapshot.
  - File: `src/app/api/cron/airdrop-price/route.ts:22`
  - Suggestion: Either expose `GET` for the cron handler to match the existing Vercel cron behavior in this repo, or add a shared handler used by both `GET` and `POST` if manual POST triggering is still desired.

### Decision
Requesting changes because the daily snapshot cron will not run successfully in production with the current method/export setup.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The follow-up update resolves the cron invocation issue from my prior review. The scheduled route now exports `GET`, which matches the Vercel cron registration in `vercel.json`.

### Findings
- No remaining code-level blockers in the changed files.

### Decision
Approving because the daily price snapshot cron I reviewed now matches issue #890 and the registered Vercel cron behavior.

PR #941, opened 2026-04-22
[#938] Fix chart: empty state, projection, mobile overflow

### LINKED ISSUE #938: [Airdrop] Fix chart: missing lines, incorrect projection, mobile responsiveness
Parent: #877

### 1. Chart shows no lines except linear projection
The "Actual FDV" and "Pool value" lines are missing. Only the dashed linear projection is visible.

**Root cause:** The chart reads from `/api/airdrop/daily-prices` which queries `pl_daily_prices`. Since the campaign hasn't started and the daily price cron hasn't run, there's no data — so `actualFdvPath` and `poolStepPath` are empty strings.

**Fix:** When there's no daily price history yet, show at minimum:
- A single point for current FDV (from the status API `currentFdv`) at today's position on the timeline
- The pool value line should show $0 flat line until now (pre-Bronze zone)
- The heartbeat dot should still appear at the current FDV position

The chart should never look empty — even before campaign start, show the current state.

### 2. Linear projection line seems incorrect
The projection starts from the bottom-left and goes to top-right, but it should start from the **current FDV at today's date** and project linearly to the Diamond target ($100M) at month 6. Currently it looks like it starts from month 1 regardless of where we are in the timeline.

**Fix:** Ensure the projection line starts from `(nowX, currentFdvY)` — not from the campaign start.

### 3. Mobile horizontal scroll
The chart overflows on mobile, causing a horizontal scroll bar. The SVG has `minWidth: 320` but the labels and axis text don't scale down properly.

**Fix options:**
- Remove `minWidth: 320` — let SVG scale freely with `viewBox`
- On mobile (< 640px): hide axis labels, increase font sizes, simplify to show only milestone emoji markers
- Or: replace SVG chart with a simplified mobile layout (milestone cards stacked vertically with progress indicators)

## Files

- `src/components/airdrop/CampaignHero.tsx` — TimelineChart component

## Acceptance Criteria

- [ ] Chart shows current FDV point even with no daily price history
- [ ] Pool value line visible (even if $0 flat line pre-Bronze)
- [ ] Linear projection starts from current position, not campaign start
- [ ] Heartbeat dot always visible at current FDV
- [ ] No horizontal scroll on mobile
- [ ] Chart text/labels readable on mobile

realproject7:
## Summary
- Chart now shows current FDV point and pool value line even with no daily price history (empty `pl_daily_prices`), so the chart never looks empty
- Heartbeat dot always visible at current FDV position regardless of historical data
- Removed `minWidth: 320` that caused horizontal scroll on mobile; axis labels and Y-left ticks hidden on small screens via Tailwind responsive classes

Fixes #938

## Test plan
- [ ] Verify chart shows current FDV dot + $0 pool line with no daily price data
- [ ] Confirm heartbeat dot pulses at current FDV even without history
- [ ] Check projection line starts from current time/FDV position
- [ ] Test on mobile viewport — no horizontal scroll
- [ ] Verify axis labels hidden on small screens, visible on desktop
- [ ] With daily price data present, chart renders all 3 lines as before

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "plotlink",
-  "version": "0.1.37",
+  "version": "0.1.38",
   "private": true,
   "workspaces": [
     "packages/*"
--- a/src/components/airdrop/CampaignHero.tsx
+++ b/src/components/airdrop/CampaignHero.tsx
@@ -180,7 +180,7 @@ function TimelineChart({
   const endMs = new Date(campaignEnd + "T00:00:00Z").getTime();
   const totalMs = endMs - startMs;
 
-  const nowX = timeToX(Math.min(nowMs, endMs), startMs, totalMs);
+  const nowX = timeToX(Math.max(startMs, Math.min(nowMs, endMs)), startMs, totalMs);
 
   const diamondFdv = tiers[tiers.length - 1].fdv;
   const fdvLogMax = Math.log10(diamondFdv * 2); // 2x headroom above diamond
--- a/src/components/airdrop/CampaignHero.tsx
+++ b/src/components/airdrop/CampaignHero.tsx
@@ -201,12 +201,24 @@ function TimelineChart({
     return result;
   }, [campaignStart, endMs]);
 
-  // Pool value step line from daily price data
+  // Effective data points: use daily prices if available, else synthesize from current FDV
+  const hasHistory = !!(dailyPrices?.length && dailyPrices.some((dp) => {
+    const dpMs = new Date(dp.date + "T00:00:00Z").getTime();
+    return dpMs >= startMs && dpMs <= endMs;
+  }));
+
+  // Pool value step line from daily price data (or $0 flat line if no data)
   const poolStepPath = useMemo(() => {
-    if (!dailyPrices?.length) return "";
+    const baseline = poolToY(0, yLeftMax);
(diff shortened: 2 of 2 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The empty-state changes are close, but there is still a correctness bug in the timeline positioning before the campaign starts.

### Findings
- [high] Empty-state current point can render left of the chart before campaign start
  - File: `src/components/airdrop/CampaignHero.tsx:183`
  - Suggestion: clamp the time used for `nowX` to the campaign window on both ends, not just the upper bound. Right now `timeToX(Math.min(nowMs, endMs), ...)` allows dates before `campaignStart`, so before kickoff the synthesized point/heartbeat/pool path can be drawn left of `PAD.left` and partially disappear. Use something like `Math.max(startMs, Math.min(nowMs, endMs))`.

### Decision
Requesting changes because the new acceptance criterion says the chart should still show the current state even before there is daily history, and this case still breaks before the campaign window begins.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The requested timeline clamp fix is in place, and the empty-state/mobile chart updates now satisfy the issue requirements.

### Findings
- None.

### Decision
Approving because the prior blocking issue was resolved and the required checks are now passing.

Issue #958, opened 2026-04-23
[Feature] Active story badges + active-first sorting on Trending/MCap


realproject7:
## Overview

Make active stories (still accepting new plots) visually distinct and prioritized in relevant sort modes.

## 1. Active badge on book card

Add an "Active" badge on the book-shaped story card, positioned **above** the plot count badge at the bottom.

```
┌─────────────┐
│ MYSTERY     │  ← genre badge (existing)
│             │
│  Story      │
│  Title      │
│             │
│ Active      │  ← new badge, same style as genre
│ 3 plots     │  ← existing plot count
└─────────────┘
```

**Badge design:**
- Same style as the genre/category badge (small rounded pill, point color text)
- Text: "Active" for active stories, "Completed" or "Expired" for sunset/expired
- Color: accent/brown for Active, muted/grey for Completed

**Book outline:**
- Active stories: thicker border with accent color (e.g., `border-2 border-accent`)
- Completed stories: normal thin border (existing style)

**Active = `sunset === false AND (has_deadline === false OR deadline not expired)`**

## 2. Active-first sorting (Trending/MCap only)

Active stories sort above completed stories, but **only** for these sort modes:
- **Trending** — active stories first, then completed
- **Market Cap** (#957) — active stories first, then completed

These sorts should NOT apply active-first:
- **Recent** — pure chronological, active status irrelevant
- **TVL** — value-based, expired stories with high TVL deserve their rank

Implementation: for Trending/MCap sorts, use a two-level sort:
```sql
ORDER BY 
  (CASE WHEN sunset = false AND (has_deadline = false OR deadline_not_expired) THEN 0 ELSE 1 END),
  mcap DESC  -- or trending metric
```

## 3. Completed/Expired visual dimming

Completed stories on the discover page:
- Book card has slightly reduced opacity or greyed-out tone
- "Completed" badge in muted color (replaces "Active" position)
- No accent border (normal thin border)

## Files

- Story card component (book shape) — add Active/Completed badge + border styling
- Discover page — add active-first logic for Trending/MCap sorts
- API route for discovery — support active-first ordering

## Acceptance Criteria

- [ ] "Active" badge on book card (same style as genre badge)
- [ ] "Completed"/"Expired" badge for sunset/expired stories
- [ ] Active stories have thicker accent-colored border
- [ ] Trending/MCap sorts: active stories first, then completed
- [ ] Recent/TVL sorts: unaffected by active status
- [ ] Completed stories visually dimmed

PR #969, opened 2026-04-23
[#957] Add Market Cap sort option

### LINKED ISSUE #957: [Feature] Add Market Cap sort option to story discovery
## Problem

Stories on the discover/home page can only be sorted by default order. Users who want to find trending or high-value stories have no way to sort by market cap.

## Solution

Add a "Market Cap" option to the sort dropdown on the story discovery page. Stories are sorted by their token's market cap (total supply × price per token × PLOT USD price) in descending order.

## Implementation

- The `trade_history` table already has `price_per_token` and `total_supply` for each storyline token
- PLOT USD price is available via `getPlotUsdPrice()`
- MCap = `total_supply × price_per_token × PLOT_USD`
- Add sort option to existing dropdown (alongside "Newest", "Most Plots", etc.)

## Files

- Discover/home page component — add sort option
- API route for discovery — add `sort=mcap` query param support

## Acceptance Criteria

- [ ] "Market Cap" sort option in dropdown
- [ ] Stories sorted by MCap descending
- [ ] Works on both desktop and mobile

realproject7:
## Summary
- Adds "Market Cap" sort option to the FilterBar dropdown (alongside Recent and Trending)
- New `getMcapStorylines()` in `lib/ranking.ts` computes MCap as `totalSupply × pricePerToken × PLOT/USD` using batch multicall + cached USD price
- MCap sort uses active-first ordering (consistent with Trending per #958)
- Reuses existing `fetchCandidatesAndRatings` for candidate pool and `getBatchTokenData` for on-chain data

Fixes #957

## Test plan
- [ ] "Market Cap" appears in sort dropdown on desktop and mobile
- [ ] Selecting Market Cap sorts stories by MCap descending
- [ ] Active stories appear above completed/expired in MCap sort
- [ ] Recent and Trending sorts unaffected
- [ ] Works with writer/genre/language filters applied

### DIFF
--- a/lib/ranking.ts
+++ b/lib/ranking.ts
@@ -1,5 +1,6 @@
 import { type Address, formatUnits } from "viem";
-import { get24hPriceChange, getTokenTVL } from "./price";
+import { get24hPriceChange, getTokenTVL, getBatchTokenData } from "./price";
+import { getPlotUsdPrice } from "./usd-price";
 import { STORY_FACTORY } from "./contracts/constants";
 import type { Database, Storyline, User } from "./supabase";
 import type { SupabaseClient } from "@supabase/supabase-js";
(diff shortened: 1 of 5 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new Market Cap tab wiring is straightforward, but the ranking source is not actually global market-cap order. `getMcapStorylines()` reuses the recency-biased candidate pool from `fetchCandidatesAndRatings()`, so older high-cap stories can be omitted entirely from the sorted results.

### Findings
- [high] Market Cap sorting is computed over a limited recency/activity subset instead of the full discover set. `getMcapStorylines()` calls `fetchCandidatesAndRatings()`, and that helper only pulls up to 50 newest stories plus 50 recently-updated stories before sorting. As a result, a story with a top market cap can be missing from the Market Cap view if it is older and not recently active, which violates the requirement that stories be sorted by MCap descending.
  - File: `lib/ranking.ts:293`
  - Suggestion: Source the MCap ranking from the full filtered discover set (or a dedicated MCap candidate query), not the trending candidate pool optimized for recency/activity.

### Decision
Requesting changes because the new sort order is incomplete and can return the wrong top stories for Market Cap.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review update fixes the blocking Market Cap ranking issue by sourcing candidates from the full filtered story set instead of the recency/activity subset used for trending. The new sort option, page wiring, and active-first ordering now line up with the reviewed requirements.

### Findings
- No blocking findings.

### Decision
Approving because the previous correctness issue in `lib/ranking.ts` is resolved and the Market Cap view now ranks the full eligible set by MCap descending with the intended active-first ordering. Checks visible to me were Vercel-only at review time.

PR #971, opened 2026-04-23
[#963] Add txHash deduplication to indexer endpoints

### LINKED ISSUE #963: [Security] Add txHash deduplication cache to prevent replay abuse
## Problem

The same txHash can be submitted unlimited times within the 5-minute validation window. Each submission triggers RPC calls and IPFS fetches. The DB upsert deduplicates the data, but the compute cost is wasted.

## Fix

Check the database for existing records before re-processing. **Do NOT use in-memory cache — plotlink runs on Vercel serverless where in-memory state doesn't persist between invocations.**

### Recommended approach: DB-first check

Before doing any RPC/IPFS work, check if the txHash is already indexed:

```typescript
// At the start of the handler, before RPC calls:
const { data: existing } = await supabase
  .from("storylines")  // or "plots", "trade_history", "donations"
  .select("id")
  .eq("tx_hash", txHash)
  .limit(1)
  .single();

if (existing) {
  return NextResponse.json({ ok: true, cached: true });
}
```

This is a fast DB query (indexed column) that prevents redundant RPC + IPFS processing.

### Per-endpoint check:

| Endpoint | Table to check | Column |
|----------|---------------|--------|
| `/api/index/storyline` | `storylines` | `tx_hash` (may need to check `plots` for genesis) |
| `/api/index/plot` | `plots` | `tx_hash` |
| `/api/index/trade` | `trade_history` | `tx_hash` + `log_index` |
| `/api/index/donation` | `donations` | `tx_hash` |

### Do NOT use:

- In-memory `Map` or `Set` — unreliable on Vercel serverless

## Impact Analysis

**plotlink-ows:** ⚠️ CRITICAL — `indexWithDelayAndRetry()` resubmits the same txHash up to 10 times when indexing fails.

**Compatibility rules:**
1. **Successfully indexed → return `{ ok: true, cached: true }`** — plotlink-ows sees success, stops retrying. Correct.
2. **Not yet indexed → process normally** — first attempt or previous attempts failed. Retries work.
3. **Never return an error for a successfully indexed txHash** — plotlink-ows would keep retrying uselessly.

**Why this is safe:** If the first indexing attempt succeeded but the client didn't get the response (network error), the retry hits the DB check, finds the record, returns `{ ok: true, cached: true }`. plotlink-ows stops retrying. Correct behavior.

**Backfill cron:** Uses direct DB queries and its own cursor. Not affected.

## Files

- `src/app/api/index/storyline/route.ts` — add DB check at start
- `src/app/api/index/plot/route.ts` — add DB check at start
- `src/app/api/index/trade/route.ts` — add DB check at start
- `src/app/api/index/donation/route.ts` — add DB check at start

## Acceptance Criteria

- [ ] DB check before RPC/IPFS processing on all 4 endpoints
- [ ] Already-indexed txHash returns 200 `{ ok: true, cached: true }`
- [ ] Not-yet-indexed txHash processes normally
- [ ] No in-memory caching (serverless-safe)
- [ ] plotlink-ows retry loop works correctly

realproject7:
## Summary
- Adds DB-first dedup check at the top of all 4 indexer endpoints (storyline, plot, donation, trade)
- Already-indexed txHash returns `200 { ok: true, cached: true }` — skips all RPC/IPFS work
- Not-yet-indexed txHash processes normally
- No in-memory caching — serverless-safe (Vercel compatible)
- Compatible with plotlink-ows retry loop: successful index on retry returns cached success

Fixes #963

## Test plan
- [ ] First submission of a txHash processes normally
- [ ] Second submission of same txHash returns `{ ok: true, cached: true }`
- [ ] No RPC/IPFS calls made on duplicate submission
- [ ] plotlink-ows retry flow works correctly (cached response stops retries)
- [ ] All 4 endpoints have dedup check

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "plotlink",
-  "version": "0.1.50",
+  "version": "0.1.51",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "plotlink",
-      "version": "0.1.50",
+      "version": "0.1.51",
       "workspaces": [
         "packages/*"
       ],
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "plotlink",
-  "version": "0.1.50",
+  "version": "0.1.51",
   "private": true,
   "workspaces": [
     "packages/*"
--- a/src/app/api/index/donation/route.ts
+++ b/src/app/api/index/donation/route.ts
@@ -28,6 +28,20 @@ export async function POST(req: Request) {
     return error("Missing or invalid txHash");
   }
 
+  // 0. DB dedup — skip RPC/IPFS if already indexed
+  const supabaseDedup = createServerClient();
+  if (supabaseDedup) {
+    const { data: existing } = await supabaseDedup
+      .from("donations")
+      .select("id")
+      .eq("tx_hash", txHash)
+      .limit(1)
+      .maybeSingle();
+    if (existing) {
+      return NextResponse.json({ ok: true, cached: true });
+    }
+  }
+
   // 1. Validate tx exists and is recent (< 5 min) — prevents spam
   const receipt = await validateRecentTx(txHash);
   if (!receipt) {
--- a/src/app/api/index/plot/route.ts
+++ b/src/app/api/index/plot/route.ts
@@ -34,6 +34,20 @@ export async function POST(req: Request) {
     return error("Missing or invalid txHash");
   }
 
+  // 0. DB dedup — skip RPC/IPFS if already indexed
+  const supabaseDedup = createServerClient();
+  if (supabaseDedup) {
+    const { data: existing } = await supabaseDedup
+      .from("plots")
+      .select("id")
+      .eq("tx_hash", txHash)
+      .limit(1)
+      .maybeSingle();
+    if (existing) {
+      return NextResponse.json({ ok: true, cached: true });
+    }
+  }
+
   // 1. Validate tx exists and is recent (< 5 min) — prevents spam
   const receipt = await validateRecentTx(txHash);
   if (!receipt) {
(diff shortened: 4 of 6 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The DB-first dedup pattern is fine for single-row endpoints, but the trade endpoint can index multiple rows per tx. The new early-return check treats any existing `tx_hash + token_address` row as a fully cached success, which can mask partial indexing and drop remaining trade logs on retries.

### Findings
- [high] Trade dedup can incorrectly short-circuit after a partial insert. `src/app/api/index/trade/route.ts` now returns `{ ok: true, cached: true }` if it finds any existing row for the submitted `tx_hash` and `token_address`. But the handler can upsert multiple trade rows for one tx (different `log_index` values). If the first attempt inserted one row and then failed before indexing the rest, a retry would hit this dedup check and skip the remaining logs forever. That loses trade history instead of just avoiding duplicate work.
  - File: `src/app/api/index/trade/route.ts:29`
  - Suggestion: Dedup trade retries at the per-log level, or only return cached success when the receipt's expected relevant logs are all already present (for example by comparing the set/count of matching `log_index` values).

### Decision
Requesting changes because the current trade dedup logic can turn a retriable partial failure into permanent missing indexed rows.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review update resolves the blocking trade-indexer issue by removing the coarse early dedup from `src/app/api/index/trade/route.ts`, so retries can still recover partial multi-log indexing failures. The remaining DB-first dedup checks are confined to the 1:1 endpoints where they correctly avoid redundant RPC/IPFS work.

### Findings
- No blocking findings.

### Decision
Approving because the previous correctness issue in the trade route is fixed and the dedup strategy now matches the data shape of each indexer endpoint. Checks visible to me were still pending at review time.

PR #972, opened 2026-04-23
[#964] Add IPFS response size limit

### LINKED ISSUE #964: [Security] Add IPFS response size limit to prevent memory exhaustion
## Problem

The storyline and plot indexer endpoints fetch content from IPFS with only a 10-second timeout — no size limit. A malicious IPFS CID could serve a multi-GB response, exhausting server memory.

## Fix

Add response size validation before reading the body:

```typescript
const ipfsRes = await fetch(ipfsUrl, { signal: AbortSignal.timeout(10_000) });
if (!ipfsRes.ok) throw new Error("IPFS fetch failed");

const contentLength = ipfsRes.headers.get("content-length");
if (contentLength && parseInt(contentLength) > 1_000_000) {
  throw new Error("IPFS content too large");
}

const text = await ipfsRes.text();
if (text.length > 1_000_000) {
  throw new Error("IPFS content too large");
}
```

1MB limit is generous — story content has a 10K char limit, so even with JSON wrapper + metadata, it should never exceed 100KB.

## Impact Analysis

**plotlink-ows:** Safe. Story content is capped at 10K chars. IPFS uploads are `JSON.stringify({ title, genre, content })` — well under 1MB.

**CLI:** Safe. Same content limits.

**Backfill cron:** Same IPFS fetch pattern — should also get the size check. Check `src/app/api/cron/backfill/route.ts`.

## Files

- `src/app/api/index/storyline/route.ts` — IPFS fetch
- `src/app/api/index/plot/route.ts` — IPFS fetch
- `src/app/api/cron/backfill/route.ts` — may need same fix

## Acceptance Criteria

- [ ] IPFS responses > 1MB rejected
- [ ] Check both Content-Length header and actual body size
- [ ] Applied to storyline, plot indexers and backfill cron
- [ ] Normal story publishing unaffected

realproject7:
## Summary
- Adds 1MB size limit on IPFS responses in storyline, plot indexers and backfill cron
- Checks both `Content-Length` header (fast reject) and actual body length (defense in depth)
- Oversized responses throw/return null — fallback content handling takes over
- Normal story content (~10K chars) is well under the 1MB limit

Fixes #964

## Test plan
- [ ] Normal story publishing unaffected (content well under 1MB)
- [ ] IPFS responses > 1MB are rejected before full body read (Content-Length check)
- [ ] IPFS responses > 1MB without Content-Length header are caught after read
- [ ] Backfill cron also rejects oversized IPFS content
- [ ] Fallback content still works when IPFS fetch fails due to size

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "plotlink",
-  "version": "0.1.51",
+  "version": "0.1.52",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "plotlink",
-      "version": "0.1.51",
+      "version": "0.1.52",
       "workspaces": [
         "packages/*"
       ],
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "plotlink",
-  "version": "0.1.51",
+  "version": "0.1.52",
   "private": true,
   "workspaces": [
     "packages/*"
--- a/src/app/api/cron/backfill/route.ts
+++ b/src/app/api/cron/backfill/route.ts
@@ -13,6 +13,7 @@ import type { Database } from "../../../../../lib/supabase";
 
 const IPFS_GATEWAY = "https://ipfs.filebase.io/ipfs/";
 const IPFS_TIMEOUT_MS = 10_000;
+const IPFS_MAX_BYTES = 1_000_000;
 
 /**
  * How many blocks to scan per cron run (~5 min on Base = ~150 blocks at 2s/block).
@@ -36,7 +37,11 @@ async function fetchIPFSContent(cid: string): Promise<string | null> {
       signal: AbortSignal.timeout(IPFS_TIMEOUT_MS),
     });
     if (!res.ok) return null;
-    return await res.text();
+    const cl = res.headers.get("content-length");
+    if (cl && parseInt(cl) > IPFS_MAX_BYTES) return null;
+    const text = await res.text();
+    if (text.length > IPFS_MAX_BYTES) return null;
+    return text;
   } catch {
     return null;
   }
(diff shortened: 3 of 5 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new `Content-Length` guard is good, but the fallback body-size check is measuring JavaScript string length rather than response bytes. That means a UTF-8 response with multi-byte characters can exceed the intended 1MB limit and still pass.

### Findings
- [medium] The post-read size validation is not actually checking bytes. `ipfsContent.length` / `text.length` counts UTF-16 code units, not the number of bytes received over the network. A malicious IPFS response containing multi-byte UTF-8 characters can be larger than 1,000,000 bytes while still staying under the character-count threshold, so the "actual body size" defense-in-depth check is incomplete.
  - File: `src/app/api/index/storyline/route.ts:126`, `src/app/api/index/plot/route.ts:98`, `src/app/api/cron/backfill/route.ts:42`
  - Suggestion: Measure bytes instead, for example with `new TextEncoder().encode(text).length` or by enforcing the limit while streaming the response body.

### Decision
Requesting changes because the current implementation can still allow oversized IPFS bodies through when the response uses multi-byte characters.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review update fixes the body-size validation by measuring UTF-8 byte length with `TextEncoder` rather than string length, so the 1MB cap now applies correctly to multi-byte IPFS responses as well. The header check plus byte-count fallback are in place for storyline, plot, and backfill paths.

### Findings
- No blocking findings.

### Decision
Approving because the previous correctness issue in the IPFS size-limit enforcement is resolved and the PR now satisfies the reviewed acceptance criteria. Checks visible to me were still pending at review time.

PR #973, opened 2026-04-23
[#965] Add fallback content size limit

### LINKED ISSUE #965: [Security] Add fallback content size limit on indexer endpoints
## Problem

The storyline and plot indexer endpoints accept a `content` field as fallback when IPFS fetch fails. This field has no size validation — an attacker could send a massive payload.

## Fix

Validate `fallbackContent` size before use:

```typescript
if (body.content && typeof body.content === "string" && body.content.length > 50_000) {
  return error("Fallback content too large", 400);
}
```

50KB is generous — plot content is capped at 10K chars.

## Impact Analysis

**plotlink-ows:** Safe. Content is already limited to 10K chars in the UI. The fallback content sent during indexing is the raw story text, well under 50KB.

**CLI:** Safe. Same content limits.

## Files

- `src/app/api/index/storyline/route.ts`
- `src/app/api/index/plot/route.ts`

## Acceptance Criteria

- [ ] Fallback content > 50KB rejected with 400 error
- [ ] Normal publish flow unaffected

realproject7:
## Summary
- Adds 50KB size limit on `body.content` fallback field in storyline and plot indexer endpoints
- Returns 400 error for oversized payloads before any processing
- Normal content (capped at 10K chars in UI) is well under the limit

Fixes #965

## Test plan
- [ ] Normal story publishing with fallback content unaffected
- [ ] Fallback content > 50KB rejected with 400
- [ ] Storyline and plot endpoints both validate

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "plotlink",
-  "version": "0.1.52",
+  "version": "0.1.53",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "plotlink",
-      "version": "0.1.52",
+      "version": "0.1.53",
       "workspaces": [
         "packages/*"
       ],
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "plotlink",
-  "version": "0.1.52",
+  "version": "0.1.53",
   "private": true,
   "workspaces": [
     "packages/*"
--- a/src/app/api/index/plot/route.ts
+++ b/src/app/api/index/plot/route.ts
@@ -30,7 +30,7 @@ export async function POST(req: Request) {
   const body = await req.json();
   const txHash = body.txHash as Hex | undefined;
   const fallbackContent = body.content as string | undefined;
-  if (fallbackContent && fallbackContent.length > 50_000) {
+  if (fallbackContent && new TextEncoder().encode(fallbackContent).byteLength > 50_000) {
     return error("Fallback content too large", 400);
   }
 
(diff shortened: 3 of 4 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new limit is on the right field, but the implementation only checks string length, not actual request size in bytes. That leaves the same multi-byte UTF-8 bypass as the IPFS body-size bug from the previous ticket.

### Findings
- [medium] Fallback content size validation uses `fallbackContent.length`, which counts JavaScript string units rather than payload bytes. A multi-byte UTF-8 request body can exceed the intended 50KB limit while still staying under 50,000 characters, so oversized fallback payloads can still get through.
  - File: `src/app/api/index/storyline/route.ts:36`, `src/app/api/index/plot/route.ts:33`
  - Suggestion: Measure bytes with `new TextEncoder().encode(fallbackContent).byteLength` (or an equivalent byte-based limit) before accepting the fallback content.

### Decision
Requesting changes because the current guard does not reliably enforce the intended 50KB payload limit.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review update fixes the fallback-content limit by measuring UTF-8 byte length with `TextEncoder` instead of string length, so the 50KB cap now applies correctly to multi-byte payloads. The check is in place for both storyline and plot indexer endpoints.

### Findings
- No blocking findings.

### Decision
Approving because the previous correctness issue in the fallback-content size validation is resolved and the PR now satisfies the reviewed acceptance criteria. Checks visible to me were Vercel-only at review time.

PR #974, opened 2026-04-23
[#966] Airdrop endpoint security: rate limiting + cache headers

### LINKED ISSUE #966: [Security] Review airdrop API endpoints for abuse vectors
## Problem

The airdrop endpoints are public and could be abused.

### Endpoints to secure:

| Endpoint | Method | Risk |
|----------|--------|------|
| /api/airdrop/checkin | POST | SIWE verified, but no rate limit. Spam failed attempts. |
| /api/airdrop/referral-code | POST | Generates codes. No rate limit. |
| /api/airdrop/register-referral | POST | Records referral. Could spam with fake wallet pairs. |
| /api/airdrop/status | GET | Calls getPlotUsdPrice() — could exhaust price API limits. |
| /api/airdrop/points | GET | Read-only DB query. |
| /api/airdrop/leaderboard | GET | Heavy query (SUM + GROUP BY). DoS via rapid requests. |

### POST endpoints: Rate limiting

Use the same rate limiting approach as #962 (DB-based or Vercel KV — NOT in-memory on serverless).

- `/api/airdrop/checkin` — 5/min per IP (SIWE is compute-heavy)
- `/api/airdrop/referral-code` — 5/min per IP
- `/api/airdrop/register-referral` — 5/min per IP

### GET endpoints: Response caching

Add `Cache-Control` headers to reduce repeated compute:

- `/api/airdrop/status` — `s-maxage=60` (1 min)
- `/api/airdrop/leaderboard` — `s-maxage=30` (30 sec)
- `/api/airdrop/points` — `s-maxage=10` (10 sec, user-specific)
- `/api/airdrop/snapshots` — `s-maxage=300` (5 min, rarely changes)

### Leaderboard query optimization

If `pl_points` grows large, the leaderboard query becomes expensive. Consider:
- Materialized view or summary table updated by the points cron
- Or at minimum, cache the response server-side for 30 seconds

## Impact Analysis

**plotlink-ows:** Not affected — doesn't call airdrop endpoints.

## Files

- All airdrop API routes in `src/app/api/airdrop/*/route.ts`
- Share rate limit utility with #962

## Acceptance Criteria

- [ ] POST endpoints rate limited (5/min per IP)
- [ ] GET endpoints have Cache-Control headers
- [ ] Rate limiting persists across serverless invocations (NOT in-memory)
- [ ] Leaderboard query doesn't degrade under load

realproject7:
## Summary
- **POST rate limiting** (5/min per IP) on checkin, referral-code, register-referral endpoints
  - DB-based via new `rate_limits` table — serverless-safe (no in-memory state)
  - Auto-purge trigger removes entries older than 5 minutes
  - Shared `lib/rate-limit.ts` utility with `checkRateLimit()` and `getClientIp()`
- **GET cache headers** on status (60s), leaderboard (30s), points (10s), snapshots (300s)
  - Uses `s-maxage` + `stale-while-revalidate` for CDN-level caching
- Migration: `00037_rate_limits.sql` creates table with index + purge trigger

Fixes #966

## Test plan
- [ ] POST endpoints return 429 after 5 requests/min from same IP
- [ ] Rate limiting persists across serverless invocations (DB-based)
- [ ] GET endpoints include Cache-Control headers in responses
- [ ] Normal airdrop flows (checkin, referral, points) unaffected
- [ ] Migration runs cleanly on Supabase

### DIFF
--- a/lib/rate-limit.ts
+++ b/lib/rate-limit.ts
@@ -1,0 +1,39 @@
+import { createServerClient } from "./supabase";
+
+/**
+ * DB-based rate limiter (serverless-safe).
+ * Returns true if the request is allowed, false if rate limited.
+ */
+export async function checkRateLimit(
+  ip: string,
+  endpoint: string,
+  maxRequests = 5,
+  windowMs = 60_000,
+): Promise<boolean> {
+  const supabase = createServerClient();
+  if (!supabase) return true; // fail open if DB unavailable
+
+  const key = `${endpoint}:${ip}`;
+  const windowStart = new Date(Date.now() - windowMs).toISOString();
+
+  // Count recent requests in window
+  const { count } = await supabase
+    .from("rate_limits")
+    .select("id", { count: "exact", head: true })
+    .eq("key", key)
+    .gte("created_at", windowStart);
+
+  if (count !== null && count >= maxRequests) {
+    return false;
+  }
+
+  // Record this request
+  await supabase.from("rate_limits").insert({ key });
+
+  return true;
+}
+
+/** Extract client IP from request headers (Vercel x-forwarded-for). */
+export function getClientIp(req: Request): string {
+  return req.headers.get("x-forwarded-for")?.split(",")[0]?.trim() ?? "unknown";
+}
--- a/lib/rate-limit.ts
+++ b/lib/rate-limit.ts
@@ -1,7 +1,8 @@
 import { createServerClient } from "./supabase";
 
 /**
- * DB-based rate limiter (serverless-safe).
+ * DB-based rate limiter (serverless-safe, atomic).
+ * Uses a Supabase RPC function that counts + inserts in a single transaction.
  * Returns true if the request is allowed, false if rate limited.
  */
 export async function checkRateLimit(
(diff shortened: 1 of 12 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The new rate-limiting utility is serverless-safe, but it is not concurrency-safe. Because it does a separate `count` query and then an `insert`, concurrent requests from the same IP can all observe the same pre-insert count and exceed the intended 5/min limit.

### Findings
- [high] `checkRateLimit()` is race-prone and can be bypassed under concurrent load. The current implementation does `SELECT count(*)` for the window and then performs a separate `INSERT`. If multiple requests from the same IP hit at once, they can all see `count < 5` before any of them insert, so more than 5 requests/min will be accepted. For a security control, this is a real enforcement bug.
  - File: `lib/rate-limit.ts:18`
  - Suggestion: Move this into a single atomic database operation, such as an RPC/function that both checks and records in one transaction, or a locking strategy that prevents concurrent over-admission.

### Decision
Requesting changes because the rate limit can be bypassed by concurrency, so the main security control in this PR is not reliably enforced.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
Moving the rate limit into a Supabase RPC is the right fix conceptually, but the new SQL function is not valid as written. It tries to apply `FOR UPDATE` to an aggregate `count(*)` query, which PostgreSQL does not allow, so the migration/function will fail or be unusable.

### Findings
- [high] `check_rate_limit()` uses `SELECT count(*) ... FOR UPDATE`, but `FOR UPDATE` cannot be applied to aggregate queries in PostgreSQL. That means the new atomic rate limiter does not actually deploy/work in its current form.
  - File: `supabase/migrations/00037_rate_limits.sql:34`
  - Suggestion: Use a transaction-safe locking strategy that locks rows or a per-key coordinator row before counting/inserting, or restructure the function so it does not rely on `FOR UPDATE` over an aggregate result.

### Decision
Requesting changes because the new database-side rate-limit function is invalid as written, so the main security control in this PR still is not ready.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The re-review update fixes the database-side rate limiter by replacing the invalid aggregate `FOR UPDATE` approach with a transaction-scoped advisory lock per key. That gives the POST airdrop endpoints a valid atomic rate-limit check while keeping the GET cache headers in place.

### Findings
- No blocking findings.

### Decision
Approving because the previous SQL validity issue in `check_rate_limit()` is resolved and the PR now satisfies the reviewed security hardening requirements for the airdrop endpoints. Checks visible to me were still pending at review time.

PR #666, opened 2026-03-31
[#661] Fix hardcoded Sepolia addresses in claim and status

realproject7:
## Summary
- `claim` and `status` commands imported `MCV2_BOND_ADDRESS` and `STORY_FACTORY_ADDRESS` directly from constants, which always resolved to Sepolia addresses — causing reverts on mainnet
- Made `storyFactory` and `mcv2Bond` public readonly on the `PlotLink` client class (already resolves based on chainId)
- Updated both commands to use `client.mcv2Bond` / `client.storyFactory` instead of hardcoded imports

## Changed files
- `packages/cli/src/sdk/client.ts` — `storyFactory` and `mcv2Bond` changed from `private` to public `readonly`
- `packages/cli/src/commands/claim.ts` — use `client.mcv2Bond` instead of `MCV2_BOND_ADDRESS`
- `packages/cli/src/commands/status.ts` — use `client.mcv2Bond` and `client.storyFactory` instead of hardcoded constants

## Test plan
- [ ] `npm run build` passes
- [ ] `npm run typecheck` passes
- [ ] `plotlink claim` resolves correct Bond address on mainnet (verified during E2E in #322)
- [ ] `plotlink status` uses correct StoryFactory and Bond addresses per chain

Fixes #661

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1171, opened 2026-05-12
Add FrameProvider initialization gate and global-error boundary (#1170)

### LINKED ISSUE #1170: [Bug] Base App black screen: missing initialization gate (FrameProvider pattern)
## Problem

PlotLink shows a permanent black screen in Base App. Previous fixes (#1158, #1160, #1162, #1166, #1168) addressed symptoms but not the root cause.

## Root Cause

**PlotLink renders children immediately during provider initialization. Both working apps gate rendering until initialization completes.**

### How DropCast handles it (works)

DropCast has a `FrameProvider` that wraps all children and **blocks rendering** until SDK/wagmi initialization is done:

```tsx
// FrameProvider.tsx
if (!frameContext.isSDKLoaded) {
  return (
    <div className="min-h-screen flex items-center justify-center">
      <div className="loading-spinner" />
      <p>Loading DropCast...</p>
    </div>
  )
}
return <FrameContext.Provider value={frameContext}>{children}</FrameContext.Provider>
```

The initialization sequence:
1. WagmiProvider + RainbowKitProvider initialize
2. FrameProvider runs: SDK context check (3s timeout) → platform detection → `sdk.actions.ready()` → auto-connect
3. Sets `isSDKLoaded = true`
4. **Only then** children render

User sees: loading spinner → app content

### How mintpad handles it (works)

mintpad uses `cookieToInitialState()` to pre-hydrate wallet state from server cookies:

```tsx
// layout.tsx (server)
const cookieStore = await cookies();
<ContextProvider cookies={cookieStore.toString()}>

// ContextProvider (client)
const initialState = cookieToInitialState(wagmiAdapter.wagmiConfig, cookies);
<WagmiProvider config={config} initialState={initialState}>
```

Wagmi starts with pre-populated state → no initialization gap → children render immediately with correct data.

### What PlotLink does (broken)

PlotLink renders children **immediately** without any gate:

```tsx
// providers.tsx
<WagmiProvider config={config}>
  <QueryClientProvider>
    <RainbowKitProvider>
      {children}   ← renders immediately, providers may not be ready
    </RainbowKitProvider>
  </QueryClientProvider>
</WagmiProvider>
```

During the initialization window:
1. Wagmi, RainbowKit are still setting up
2. Children render and trigger hooks/queries before providers are ready
3. Components may throw or render empty (NavBar, Footer depend on provider state)
4. Without error boundary → React unmounts DOM → permanent **black screen** (Base App's dark webview background)

This also explains:
- **Black, not beige**: crash happens before PlotLink's CSS `body { background: var(--bg) }` takes effect
- **Permanent**: no error recovery, no error boundary
- **~75% failure rate**: race condition timing varies
- **Refresh doesn't help**: webview retains corrupted state

## Fix: Add FrameProvider pattern (recommended)

Create a `FrameProvider` component like DropCast that:
1. Wraps children inside the provider tree
2. Runs platform detection + SDK initialization
3. Shows loading indicator until ready
4. Handles auto-connect for Base App
5. Only renders children when everything is initialized

```tsx
// src/components/FrameProvider.tsx
function FrameProvider({ children }) {
  const [isReady, setIsReady] = useState(false);
  const [platform, setPlatform] = useState<Platform>("web");

  useEffect(() => {
    detectPlatform().then(p => {
      setPlatform(p);
      setIsReady(true);
    });
  }, []);

  useAutoConnectInjected(platform, isReady);

  if (!isReady) {
    return (
      <div className="min-h-screen flex items-center justify-center" 
           style={{ background: '#faf8f5' }}>
        <p>Loading PlotLink...</p>
      </div>
    );
  }

  return <>{children}</>;
}
```

Note: use inline `style={{ background: '#faf8f5' }}` for the loading state — not CSS variables — since the CSS may not be ready.

### Also add `global-error.tsx`

As a safety net, add `src/app/global-error.tsx` to catch any uncaught errors and show a recovery UI instead of a black screen.

## Architecture Comparison

| App | Initialization Gate | Loading UI | Auto-Connect | Result |
|-----|---------------------|------------|--------------|--------|
| DropCast | FrameProvider (`isSDKLoaded`) | Spinner + text | useAutoConnectInjected hook | Works |
| mintpad | cookieToInitialState (pre-hydrated) | None needed | Implicit via wagmi | Works |
| PlotLink | **None** | **None** | In ConnectWallet (no gate) | **Black screen** |

## Acceptance Criteria

- [ ] FrameProvider (or equivalent) gates child rendering until initialized
- [ ] Loading indicator shown during initialization (with inline styles, not CSS vars)
- [ ] Platform detection completes before children render
- [ ] Auto-connect fires for Base App after initialization
- [ ] `global-error.tsx` catches uncaught errors with recovery UI
- [ ] Farcaster: sdk.actions.ready(), addMiniApp, composeCast all work
- [ ] Desktop browser unchanged

Refs: #1154, #1158, #1160, #1162, #1166, #1168, dropcast FrameProvider pattern

realproject7:
## Summary
- Adds `FrameProvider` component that gates child rendering until platform detection completes, preventing race conditions that caused black screens in Base App
- Shows inline-styled loading state (no CSS variable dependency) during initialization
- Moves auto-connect logic from `ConnectWallet` into `FrameProvider` for proper sequencing — wallet connects only after platform is known
- Adds `global-error.tsx` as safety net to catch uncaught errors with a recovery UI instead of permanent black screen

## Changes
- `src/components/FrameProvider.tsx` — New initialization gate component
- `src/app/global-error.tsx` — New global error boundary
- `src/app/providers.tsx` — Wraps children with FrameProvider inside RainbowKitProvider
- `src/components/ConnectWallet.tsx` — Removes auto-connect logic (now in FrameProvider)
- `package.json` — Version bump 1.25.2 → 1.25.3

Closes #1170

## Test plan
- [ ] Base App: no black screen, loading indicator shown briefly, then app renders
- [ ] Base App: wallet auto-connects after initialization
- [ ] Farcaster: sdk.actions.ready(), addMiniApp, composeCast all work
- [ ] Desktop browser: RainbowKit connect modal works, no loading flash
- [ ] Force an error in a component — global-error.tsx shows recovery UI

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "plotlink",
-  "version": "1.25.2",
+  "version": "1.25.3",
   "private": true,
   "workspaces": [
     "packages/*"
--- a/src/app/global-error.tsx
+++ b/src/app/global-error.tsx
@@ -1,0 +1,46 @@
+"use client";
+
+export default function GlobalError({
+  reset,
+}: {
+  error: Error & { digest?: string };
+  reset: () => void;
+}) {
+  return (
+    <html lang="en">
+      <body
+        style={{
+          minHeight: "100vh",
+          display: "flex",
+          alignItems: "center",
+          justifyContent: "center",
+          background: "#faf8f5",
+          fontFamily: "system-ui, sans-serif",
+          color: "#333",
+          textAlign: "center",
+          padding: "24px",
+        }}
+      >
+        <div>
+          <h2 style={{ fontSize: "16px", marginBottom: "8px" }}>
+            Something went wrong
+          </h2>
+          <button
+            onClick={reset}
+            style={{
+              background: "#333",
+              color: "#fff",
+              border: "none",
+              borderRadius: "4px",
+              padding: "8px 16px",
+              fontSize: "13px",
+              cursor: "pointer",
+            }}
+          >
+            Try again
+          </button>
+        </div>
+      </body>
+    </html>
+  );
+}
--- a/src/app/global-error.tsx
+++ b/src/app/global-error.tsx
@@ -14,9 +14,9 @@ export default function GlobalError({
           display: "flex",
           alignItems: "center",
           justifyContent: "center",
-          background: "#faf8f5",
+          background: "#0a0a0a",
           fontFamily: "system-ui, sans-serif",
-          color: "#333",
+          color: "#999999",
           textAlign: "center",
           padding: "24px",
         }}
@@ -28,8 +28,8 @@ export default function GlobalError({
           <button
             onClick={reset}
             style={{
-              background: "#333",
-              color: "#fff",
+              background: "#00ff88",
+              color: "#0a0a0a",
               border: "none",
               borderRadius: "4px",
               padding: "8px 16px",
--- a/src/app/providers.tsx
+++ b/src/app/providers.tsx
@@ -6,6 +6,7 @@ import { RainbowKitProvider, type Theme } from "@rainbow-me/rainbowkit";
 import { config } from "../../lib/wagmi";
 import { useState, Suspense } from "react";
 import { useReferralCapture } from "../hooks/useReferralCapture";
+import { FrameProvider } from "../components/FrameProvider";
 import "@rainbow-me/rainbowkit/styles.css";
 
 // PlotLink-themed RainbowKit theme using CSS vars
@@ -103,10 +104,12 @@ export function Providers({ children }: { children: React.ReactNode }) {
             ),
           }}
         >
-          <Suspense fallback={null}>
-            <ReferralCapture />
-          </Suspense>
-          {children}
+          <FrameProvider>
+            <Suspense fallback={null}>
+              <ReferralCapture />
+            </Suspense>
+            {children}
+          </FrameProvider>
         </RainbowKitProvider>
       </QueryClientProvider>
     </WagmiProvider>
--- a/src/components/ConnectWallet.tsx
+++ b/src/components/ConnectWallet.tsx
@@ -1,12 +1,10 @@
 "use client";
 
-import { useEffect, useRef } from "react";
-import { useAccount, useConnect, useDisconnect } from "wagmi";
+import { useAccount, useDisconnect } from "wagmi";
 import { ConnectButton } from "@rainbow-me/rainbowkit";
 import Link from "next/link";
 import { truncateAddress } from "../../lib/utils";
 import { useConnectedIdentity } from "../hooks/useConnectedIdentity";
-import { usePlatformDetection } from "../hooks/usePlatformDetection";
 
 interface ConnectWalletProps {
   onNavigate?: () => void;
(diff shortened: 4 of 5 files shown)


realproject7:
## RE2 Review — ✅ Approved

Good approach — moving auto-connect into `FrameProvider` gives proper sequencing (detect → connect → render) and inline styles prevent CSS race conditions. ConnectWallet is now purely presentational.

**Notes (non-blocking):**

1. **Light loading screen vs dark theme:** `FrameProvider` and `global-error.tsx` both use `background: "#faf8f5"` (light cream). PlotLink's design system is dark (`#0a0a0a`). Users will see a light flash before the dark app renders. Consider using `background: "#0a0a0a", color: "#666"` to match the terminal aesthetic.

2. **Missing `isConnected` guard:** The auto-connect effect in `FrameProvider` doesn't check if the wallet is already connected (the previous ConnectWallet code had this guard). If wagmi's reconnect fires first, this will attempt a redundant connect. Wagmi handles it gracefully, but adding `const { isConnected } = useAccount()` and checking it would be cleaner.

3. **`connectors` in deps array:** `connectors` from `useConnect()` can change reference on re-render, potentially re-firing the auto-connect effect. Combined with no `isConnected` guard, could cause repeated attempts. Not a crash risk but worth hardening in a follow-up.

Core approach is solid. Approved.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The PR adds the requested initialization gate and inline loading UI, but the moved auto-connect logic is missing the guard that existed in `ConnectWallet`. As written, `FrameProvider` can repeatedly call `connect()` after initialization whenever hook dependencies change.

### Findings
- [high] FrameProvider auto-connect can repeat without an `isConnected` or attempted guard
  - File: `src/components/FrameProvider.tsx:24`
  - Details: The effect calls `connect()` for Base App/Farcaster whenever `isReady`, `platform`, `connectors`, or `connect` changes. Unlike the old `ConnectWallet` code, it does not check `isConnected` and does not keep an `autoConnectAttempted` ref. If wagmi returns a new `connectors` array or `connect` callback across state updates, this can repeatedly invoke `connect()` after the app is already connected or after a failed attempt. In Base App that risks repeated wallet prompts/reconnect churn during the exact startup path this ticket is trying to stabilize.
  - Suggestion: Include `useAccount().isConnected` and an attempted ref, with Base App retry behavior handled intentionally rather than on every dependency change. For example, skip when connected; mark an attempt before calling `connect`; reset only when platform changes if a retry is desired.

### Decision
Requesting changes because the new provider-level auto-connect needs the same idempotency guard as the previous implementation before it is safe to run at the root.

realproject7:
## RE2 Re-review — ✅ Approved

All feedback addressed:
- `isConnected` guard + `connectAttempted` ref prevents repeated connect calls — resolves RE1's blocker
- Dark background `#0a0a0a` on both FrameProvider loading state and global-error — no light flash
- Green accent `#00ff88` on error recovery button matches the design system

Clean fix. Approval confirmed.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The FrameProvider idempotency blocker from my previous review is fixed: the code now checks `isConnected` and uses a `connectAttempted` ref. However, the latest `e2e` check is failing, so this PR is not ready to approve.

### Findings
- [high] Latest e2e check is failing
  - File: PR checks
  - Details: `lint-and-typecheck` passes, but the latest `e2e` job failed after the FrameProvider fix commit. Because this PR changes root rendering and initialization behavior, the e2e failure needs to be investigated before merge.
  - Suggestion: Inspect the failing e2e logs, fix the regression or provide a rerun if it was infrastructure-only, then request re-review once checks are green.

### Decision
Requesting changes until the e2e failure is resolved.

realproject7:
## RE2 Follow-up — ✅ Approval confirmed

Good fix for the e2e failure. The fast-path `!window.ethereum && window.self === window.top` skips async platform detection entirely on desktop web — resolves in a single tick with no SDK import or timeout. Mini apps (injected provider or iframe) still get the full `detectPlatform()` gate.

CI pending. Approval stands.

project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The desktop fast path addresses the prior e2e timeout directionally, and the auto-connect guard remains in place. However, the latest `lint-and-typecheck` check is now failing on the newest commit, so the PR is not approvable yet.

### Findings
- [high] Latest lint/typecheck check is failing
  - File: PR checks
  - Details: After commit `f941139`, `lint-and-typecheck` fails while e2e is still pending. This PR changes root initialization code, so the type/lint failure must be fixed before merge.
  - Suggestion: Inspect the failing check logs, fix the type/lint issue, and rerun CI.

### Decision
Requesting changes until required checks are green.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The FrameProvider gate now includes an idempotent auto-connect guard and the desktop fast path no longer trips lint. The implementation matches #1170: root children are gated for mini app initialization, the loading/error UI uses inline dark styling, and desktop render remains immediate.

### Findings
- None blocking.

### Decision
Approved. Latest `lint-and-typecheck` and `e2e` checks are passing.

PR #1187, opened 2026-05-14
Add NSFW filter feedback to discover page (#1184)

### LINKED ISSUE #1184: [Bug] 18+ filter toggle has no visible effect on discover page
## Problem

The 18+ filter checkbox on the discover page appears to do nothing when toggled. The same stories are shown regardless of whether it's checked or unchecked.

## Analysis

The code logic chain is correct:
1. Checkbox → `navigate({ nsfw: true })` → URL gets `?nsfw=1`
2. Server: `showNsfw = rawNsfw === "1"` → `true`
3. Query: `if (!showNsfw) filtered.eq("is_nsfw", false)` → skipped when checked → all stories returned

**The filter works as "include 18+ content" (additive), not "show only 18+ content" (exclusive).** When checked, NSFW stories are added to the results alongside non-NSFW stories. When unchecked, NSFW stories are hidden.

### 1. No NSFW stories exist in the database
If no storylines have `is_nsfw = true`, toggling the filter produces identical results — it looks broken but technically works.

→ **Check**: Query `SELECT COUNT(*) FROM storylines WHERE is_nsfw = true` in Supabase.

### 2. NSFW stories exist but the filter effect is invisible
If only 1-2 NSFW stories exist and they sort below the visible results, the user won't notice them appearing/disappearing.

### 3. Navigation/caching issue
`router.push()` might not trigger a full re-fetch if the page is statically cached or if Next.js treats the navigation as a soft transition.

## Action Items

- [ ] Check if any NSFW stories exist in the database
- [ ] If none exist, the filter is working correctly — just no NSFW content yet
- [ ] If NSFW stories exist, verify they appear/disappear when toggling the filter
- [ ] Consider adding a "No 18+ stories found" empty state when the filter is checked but no NSFW stories exist

realproject7:
## Summary
- The 18+ toggle on the discover page appeared broken because it's additive — when no NSFW stories exist in the DB, toggling it produces identical results
- Added contextual feedback messages: "No 18+ stories found" when none exist, or "Showing N mature stories alongside regular content" when they do
- Queries NSFW count only when filter is active (no perf impact when off)

Closes #1184

## Test plan
- [ ] Toggle 18+ filter on discover page — verify feedback message appears
- [ ] If no NSFW stories exist, should show "No 18+ stories found — all current stories are safe for work"
- [ ] If NSFW stories exist, should show count of mature stories
- [ ] Verify message disappears when filter is toggled off
- [ ] Check mobile filter sheet flow works the same

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "plotlink",
-  "version": "1.25.4",
+  "version": "1.25.5",
   "private": true,
   "workspaces": [
     "packages/*"
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -35,8 +35,22 @@ export default async function Home({
   const supabase = createServerClient();
 
   let storylines: Storyline[] = [];
+  let nsfwCount = 0;
   if (supabase) {
-    storylines = await queryTab(supabase, tab, writer, page, genre, lang, showNsfw);
+    const results = await Promise.all([
+      queryTab(supabase, tab, writer, page, genre, lang, showNsfw),
+      showNsfw
+        ? supabase
+            .from("storylines")
+            .select("*", { count: "exact", head: true })
+            .eq("hidden", false)
+            .eq("is_nsfw", true)
+            .eq("contract_address", STORY_FACTORY.toLowerCase())
+            .then(({ count }) => count ?? 0)
+        : Promise.resolve(0),
+    ]);
+    storylines = results[0];
+    nsfwCount = results[1];
   }
 
   return (
@@ -51,7 +65,7 @@ export default async function Home({
         </p>
       </header>
 
-      <FilterBar writer={writer} genre={genre} lang={lang} tab={tab} totalCount={storylines.length} showNsfw={showNsfw} />
+      <FilterBar writer={writer} genre={genre} lang={lang} tab={tab} totalCount={storylines.length} showNsfw={showNsfw} nsfwCount={nsfwCount} />
 
       {/* Section label */}
       <h2 className="mt-4 mb-3 text-[11px] font-semibold uppercase tracking-[0.08em] text-muted">
--- a/src/app/page.tsx
+++ b/src/app/page.tsx
@@ -40,13 +40,20 @@ export default async function Home({
     const results = await Promise.all([
       queryTab(supabase, tab, writer, page, genre, lang, showNsfw),
       showNsfw
-        ? supabase
-            .from("storylines")
-            .select("*", { count: "exact", head: true })
-            .eq("hidden", false)
-            .eq("is_nsfw", true)
-            .eq("contract_address", STORY_FACTORY.toLowerCase())
-            .then(({ count }) => count ?? 0)
+        ? (() => {
+            let q = supabase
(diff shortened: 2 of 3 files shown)


project7-interns (changes requested):
## Verdict: REQUEST CHANGES

### Summary
The feedback UI is a reasonable direction, and CI is passing. However, the count used for the feedback does not match the discover query filters, so the page can show incorrect guidance for filtered views.

### Findings
- [medium] The `nsfwCount` query counts all visible NSFW stories for the contract, but it does not apply the active `writer`, `genre`, `lang`, or `sunset` filters used by the result query. With `?nsfw=1&genre=<genre-with-no-nsfw>` or a writer/language filter, the UI can say “Showing N mature stories alongside regular content” even when no mature stories are included in the current results. For the `new` tab it can also count sunset stories that the result query excludes.
  - File: `src/app/page.tsx:43`
  - Suggestion: Build the count from the same base filters as the active discover query. At minimum, apply the current writer/genre/lang filters and align the `hidden`, `contract_address`, and `sunset` behavior with the result set being messaged.

### Decision
Requesting changes because the added feedback can be factually wrong in common filtered discover states, which undermines the issue fix.

realproject7 (commented):
## Review — @re2 (Changes Requested)

**Issue: Missing `sunset` filter on NSFW count query**

In `src/app/page.tsx`, the new count query filters by `hidden`, `is_nsfw`, and `contract_address`, but omits `.eq("sunset", false)` which `queryTab` applies (line 134). This means the NSFW count could include sunset/deprecated stories that don't appear in results, giving users an inflated number.

```diff
  supabase
    .from("storylines")
    .select("*", { count: "exact", head: true })
    .eq("hidden", false)
+   .eq("sunset", false)
    .eq("is_nsfw", true)
    .eq("contract_address", STORY_FACTORY.toLowerCase())
```

**Minor note:** The count query doesn't apply writer/genre/lang filters, so the message "Showing N mature stories" reflects total NSFW count, not the count within the current filtered view. This is fine as general feedback but worth noting.

**Everything else looks good:**
- Version bump (1.25.4 → 1.25.5) appropriate for bug fix
- `head: true` correct for count-only queries
- `Promise.all` for parallel execution is the right pattern
- Conditional count (only when `showNsfw`) avoids unnecessary queries
- Singular/plural handling is correct
- CI passing

Please add the missing `.eq("sunset", false)` to the count query.

project7-interns (approved):
## Verdict: APPROVE

### Summary
The requested count-query fix is in place, and the feedback now aligns with the active discover filters. CI is passing on the updated commit.

### Findings
- None.

### Decision
Approved. The PR satisfies issue #1184 by making the additive 18+ filter visibly explain its effect without introducing misleading counts in filtered views.

realproject7 (commented):
## Re-review — @re2 ✓ Approved

The updated count query now correctly mirrors `queryTab` filters:
- `sunset` filter applied for the "new" tab
- Writer type, genre, and language filters all applied
- Count now reflects what's actually visible in the filtered results

CI passing. LGTM.

